PR #15, opened 2018-01-16
netstandard 2.0

LORDofDOOM:
[Core] Use Library with netstandard, move form elements (require WindowsForms-> Not available in netstandard) to own library, add compiler statement for plot


jjxtra:
Thanks for the pull request! I tweaked things a little bit and did a commit. I like the idea of keeping the forms (plot, etc.) in the library. However since these can't be available in a .net standard library, they are only available from .net 47 applications. Dual targeting allows there to be two libraries - one compatible with .net standard 2.0 without forms and another for .net 4.7 with forms. The show plot form is now an extension method on a new PlotFormExtensions in PlotForm.cs.

PR #17, opened 2018-01-21
Added market orders - implemented for binance only

luigidallavalle:


jjxtra:
I don't think this is a good idea, market orders are very dangerous. If the order book is thin or something goes wrong it can mean a lot of lost funds. For cases where you really must do an API call for a market order, you can call PlaceOrder with a very high price (buy) or low price (sell).

PR #23, opened 2018-02-07
Implement websockets for Bittrex GetTickersWebSocket

bichuga:
Using https://github.com/JKorf/Bittrex.Net to handle both the signalR interface as well as cloudflare ddos protection. (MIT license)

### DIFF
--- a/Console/ExchangeSharpConsole.cs
+++ b/Console/ExchangeSharpConsole.cs
@@ -90,6 +90,10 @@ namespace ExchangeSharpConsoleApp
                 {
                     RunProcessEncryptedAPIKeys(dict);
                 }
+                else if (dict.ContainsKey("bittrex-websocket"))
+                {
+                    RunBittrexWebSocket();
+                }
                 else
                 {
                     Console.WriteLine("Unrecognized command line arguments.");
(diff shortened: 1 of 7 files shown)


jjxtra:
Can you move the tests into the web socket test function in the console tests file? Then put the console main back to the way it was :)

bichuga:
Yes of course. Just wanted to show that this actually worked

jjxtra:
I switched to return IDisposable, more appropriate since the only purpose of the return value is to close the web socket. That should allow you to wrap Bittrex.net inside of an IDisposable, if it doesn't already implement it.

bichuga:
Alright this is ready to go.

jjxtra:
I think the only change I would want is that the ValueOrDefault calls use the ConvertInvariant<decimal> calls, to be consistant with the rest of the framework.

jjxtra:
Ok if there is nothing else I will give this a merge and quick test. Let me know!

bichuga:
Let 'er rip

jjxtra:
Thanks for doing the work for this, your time is very much appreciated. Good luck on your trading!

PR #24, opened 2018-02-08
Adding market orders to those that support it

Ofekw:


jjxtra:
Looks great, just want to verify you have tested both limit and market order on each exchange that was modified.

Ofekw:
I don't actually have access all of these. I can test Binance, Gdax and Kraken, but would need help with the others. Will comment back once verified. 

I followed their individual api docs - _we should be good_

jjxtra:
The order type names match case insensitive - I would recommend a double check and make it match case sensitive. You can use ToUpperInvariant or ToLowerInvariant after the to string for those exchanges that don't use pascal casing but have uppercased or lowercased the whole thing.

jjxtra:
Looks good now, made a few tweaks, namely Bitfinex wants "exchange limit" or "exchange market". Other than that good to go. Thanks for submitting the pull request!

Ofekw:
Thank you for this awesome library! Expect more PRs soon!

I was a little confused by ther Bitfinex api docs, cheers for that!

jjxtra:
Thank you! Looking forward to the goodies!

-- Jeff

Issue #26, opened 2018-02-10
GetOpenOrderDetails fails on Bittrex


clarity99:
I have a [BUG REPORT] 

Details:
running GetOpenOrderDetails on Bittrex fails with error System.ArgumentException: 'Can not convert Null to Decimal.'

It fails in Parse order on line: 
  order.AveragePrice = token["PricePerUnit"].ConvertInvariant<decimal>(token["Price"].ConvertInvariant<decimal>());
          
if I dump the Jobject I get this:

{
  "success": true,
  "message": "",
  "result": [
    {
      "Uuid": null,
      "OrderUuid": "6523de7d-e999-4d24-ad70-3adaf2895190",
      "Exchange": "BTC-DGB",
      "OrderType": "LIMIT_SELL",
      "Quantity": 7000.0,
      "QuantityRemaining": 7000.0,
      "Limit": 5.49E-06,
      "CommissionPaid": 0.0,
      "Price": 0.0,
      "PricePerUnit": null,
      "Opened": "2018-02-09T15:00:53.76",
      "Closed": null,
      "CancelInitiated": false,
      "ImmediateOrCancel": false,
      "IsConditional": false,
      "Condition": "NONE",
      "ConditionTarget": null
    },
    {
      "Uuid": null,
      "OrderUuid": "b6b68d04-20ff-4d4a-a108-66fd35015a8b",
      "Exchange": "BTC-DGB",
      "OrderType": "LIMIT_SELL",
      "Quantity": 5000.0,
      "QuantityRemaining": 5000.0,
      "Limit": 4.77E-06,
      "CommissionPaid": 0.0,
      "Price": 0.0,
      "PricePerUnit": null,
      "Opened": "2018-02-09T15:01:07.273",
      "Closed": null,
      "CancelInitiated": false,
      "ImmediateOrCancel": false,
      "IsConditional": false,
      "Condition": "NONE",
      "ConditionTarget": null
    },
    {
      "Uuid": null,
      "OrderUuid": "e2369e65-46be-4e9b-9812-1c9d659192ac",
      "Exchange": "BTC-DGB",
      "OrderType": "LIMIT_SELL",
      "Quantity": 2000.0,
      "QuantityRemaining": 2000.0,
      "Limit": 8E-06,
      "CommissionPaid": 0.0,
      "Price": 0.0,
      "PricePerUnit": null,
      "Opened": "2018-02-09T15:02:01.777",
      "Closed": null,
[...]

PR #27, opened 2018-02-10
PricePerUnit null fix

clarity99:
-  fix for #26 

- [ ] Fix for null PricePerUnit field from Bittrex

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeBittrexAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBittrexAPI.cs
@@ -74,7 +74,7 @@ namespace ExchangeSharp
             decimal amountFilled = amount - remaining;
             order.Amount = amount;
             order.AmountFilled = amountFilled;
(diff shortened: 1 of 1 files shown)


ExchangeSharp/API/Exchanges/ExchangeBittrexAPI.cs
```diff
@@ -74,7 +74,7 @@ private ExchangeOrderResult ParseOrder(JToken token)
             decimal amountFilled = amount - remaining;
             order.Amount = amount;
             order.AmountFilled = amountFilled;
-            order.AveragePrice = token["PricePerUnit"].ConvertInvariant<decimal>(token["Price"].ConvertInvariant<decimal>());
+            order.AveragePrice = token["PricePerUnit"].ConvertInvariant<decimal?>(token["Price"].ConvertInvariant<decimal>()) ?? 0;
```
bichuga: if token["PricePerUnit"] is null, what happens?

jjxtra:
Thanks so much for taking the time and reporting. I have fixed ConvertInvariant to handle JValue objects with null Value properties which should handle this case, and any other.

PR #28, opened 2018-02-10
Binance GetOpenOrders optimization

clarity99:
Removes the hack for Binance GetOpenOrders, as it is not needed anymore. This speeds up the call significantly (from 5 minutes to few 100 milliseconds on my machine)


jjxtra:
Thanks so much! I was wondering if they would implement this. Does the same work for /api/v3/allOrders?

clarity99:
Looking at the documentation for allOrders, it would seem that is not the case - there symbol is mandatory, whereas for openOrders it is not.

jjxtra:
Ok will merge this. In the meantime I have tweeted the heck out of their accounts to request that allOrders support the same.

PR #29, opened 2018-02-10
Time in force not relevant to Binance market orders

bichuga:


jjxtra:
Should we wrap the other market orders in this same logic?

jjxtra:
I went ahead and made the changes. Thanks for the pull request!

PR #36, opened 2018-02-15
Implementing GetOpenOrderDetails for Kraken

clausjensen:
Implementing GetOpenOrderDetails for Kraken.

Also added Price property to ExchangeOrderResult. I'm not sure where the requested price of an order would be located? .. please let me know if there's already a field for it somewhere that I've just missed.. but it seems like the only price field is the AveragePrice which is what is calculated when an order is filled.

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -531,7 +531,8 @@ namespace ExchangeSharp
             {
                 Amount = token["origQty"].ConvertInvariant<decimal>(),
                 AmountFilled = token["executedQty"].ConvertInvariant<decimal>(),
-                AveragePrice = token["price"].ConvertInvariant<decimal>(),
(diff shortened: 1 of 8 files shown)


jjxtra:
ExchangeOrderRequest has a Price field.

jjxtra:
We need to go through all the parse orders and differentiate between "Price" and "AveragePrice" which is the average execution price. Are you up for doing that?

clausjensen:
That's exactly the problem I was having ;) .. needed somewhere to put the requested trade price on the response from the API since it was only part of the request object used to create the order. Just wasn't sure if I had missed a field somewhere, where this was supposed to be put in. Are you okay with going with `.Price` then?

jjxtra:
Yea that should be fine.

clausjensen:
I don't have accounts at most of these exchanges, so I haven't been able to fully test this. I went through API docs for the exchanges where order stuff was implemented and have changed to use price instead. There's a few todos in there for implementing the average price based on fills... but someone else can have fun with that 😃

PR #38, opened 2018-02-17
[Binance] Use correct normalize

LORDofDOOM:
Binance use reversed symbol -> Normal symbol is e.g. BTC-XPM - this is XPMBTC in Binance. Not really a elegant solution but working :-)

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -34,7 +34,12 @@ namespace ExchangeSharp
         {
             if (symbol != null)
             {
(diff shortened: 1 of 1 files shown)


jjxtra:
I think this is better left to the NormalizeGlobal call. The normalize call is mainly for making sure the case and any delimiters are correct. Anyone who is using GetSymbols or passing hard-coded symbols into calls will not want the order of the pairs to get swapped.

LORDofDOOM:
I understand this problem...

If you can add e.g. NormalizeFromGlobal function where a "standard" symbol like BTC-ETH always will switch to the symbol of the current exchange (ETHBTC in this case for binance) this is the way to go :-)

jjxtra:
Going forward I think NormalieGlobal should be used. The normalize function just fixes the case and any delimiters in order to ensure the existing symbol works, it does not swap pairs.

PR #42, opened 2018-02-22
replace AES ciphermode CFB

basman:
CFB is not supported on .net core.
see also https://github.com/dotnet/corefx/issues/4647


jjxtra:
Looks like this function is not used anywhere yet so this should be just fine.

PR #43, opened 2018-02-22
RSACryptoServiceProvider does not support CspParameters on Linux

basman:
This should be tested on Windows before merging!

How I found this issue:

cd Console
dotnet run test

Fatal error: System.PlatformNotSupportedException: 'CspParameters' requires Windows Cryptographic API (CAPI), which is not available on this platform.
   at System.Security.Cryptography.RSACryptoServiceProvider..ctor(Int32 dwKeySize, CspParameters parameters)
   at ExchangeSharp.ManagedProtection.GetKey(DataProtectionScope scope) in .../ExchangeSharp-netcore/ExchangeSharp/Utility/ManagedProtection.cs:line 284
   at ExchangeSharp.ManagedProtection.Protect(Byte[] userData, Byte[] optionalEntropy, DataProtectionScope scope) in .../ExchangeSharp-netcore/ExchangeSharp/Utility/ManagedProtection.cs:line 108
   at ExchangeSharp.DataProtector.Protect(Byte[] data) in .../ExchangeSharp-netcore/ExchangeSharp/Utility/DataProtector.cs:line 131
   at ExchangeSharpConsoleApp.ExchangeSharpConsole.TestEncryption() in .../ExchangeSharp-netcore/Console/ExchangeSharpConsole_Tests.cs:line 103
   at ExchangeSharpConsoleApp.ExchangeSharpConsole.RunPerformTests(Dictionary`2 dict) in .../ExchangeSharp-netcore/Console/ExchangeSharpConsole_Tests.cs:line 171
   at ExchangeSharpConsoleApp.ExchangeSharpConsole.ConsoleMain(String[] args) in .../ExchangeSharp-netcore/Console/ExchangeSharpConsole.cs:line 67

### DIFF
--- a/ExchangeSharp/Utility/ManagedProtection.cs
+++ b/ExchangeSharp/Utility/ManagedProtection.cs
@@ -279,9 +279,16 @@ namespace ExchangeSharp
                     {
                         lock (user_lock)
                         {
-                            // CspParameters csp = new CspParameters();
-                            // csp.KeyContainerName = "DAPI";
-                            user = new RSACryptoServiceProvider(1536);
+                            if (CryptoUtility.IsMono)
+                            {
+                                CspParameters csp = new CspParameters();
+                                csp.KeyContainerName = "DAPI";
+                                user = new RSACryptoServiceProvider(1536, csp);
+                            }
+                            else
+                            {
+                                user = new RSACryptoServiceProvider(1536);
+                            }
                         }
                     }
                     return user;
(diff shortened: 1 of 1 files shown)


jjxtra:
Do we need to do the same for machine level? This was apparently working on Linux using mono, is it just in .NET core on Linux that it has the problem? If so we should add a runtime check to see if it is running under mono and keep those params...

jjxtra:
Windows is not an issue as it uses PInvoke into windows DAPI and does not used managed protection on Windows

jjxtra:
I checked in a change to CryptoUtility and added an IsMono param, if IsMono is true, we should keep the code you commented out.

basman:
I just ran the tests and user scope was the only path. I did not check if 
the tests cover machine scope. I think on Linux user scope makes most sense.

--Roman

PR #49, opened 2018-03-06
Implement GetDepositAddress for Binance/Bittrex/Bitfinex/Polo. Collect more data from GetCurrencies

bichuga:
Implemented GetDepositAddress being super careful about cryptos that require two fields to deposit. Bittrex/Polo have a crazy API that don't return the baseaddress to deposit to and rather they return the memo with no indication this is not an address. On Polo, we can query getcurrencies to see if a currency has a BaseAddress and work from there, but on Bittrex, many coins (such as BTC) have a BaseAddress which makes it impossible to programmatically query to see if a coin requires one or two fields. For now I'm keying off of coin type for Bittrex. I posted a support ticket to see if they have recommendations:
https://support.bittrex.com/hc/en-us/requests/1228941


jjxtra:
Let's change the following, let me know if you have any objections:

- Change Dictionary return types to IReadOnlyDictionary
- Add documentation comments to ExchangeDepositDetails and properties

I'll let you know if I see anything else. Thanks for the pull request!

jjxtra:
Thanks so much for putting this together!

PR #53, opened 2018-03-10
Moving average tests

basman:
Add tests for MovingAverageCalculator.
Fixed EMA weighting multiplier precision.


ExchangeSharp/Traders/MovingAverageCalculator.cs
```diff
@@ -129,7 +129,7 @@ public void Reset(int windowSize)
         {
             _windowSize = windowSize;
             _values = new double[_windowSize];
-            _weightingMultiplier = 2.0f / (_values.Length + 1);
```
basman: This unimposing change affects moving average calculation beyond the 6th fractional places. Still this can be relevant, as some exchanges use 8 fractional places.

PR #54, opened 2018-03-10
Use decimal type in MovingAverageCalculator

basman:
This one depends on the previous PR #53.

Again, this changes the exact values computed by MovingAverageCalculator.
What was your initial intention/concern using double instead? Was it performance?


jjxtra:
Do you have a specific use case where the precision gained by decimal is affecting calculations in a meaningful way?

basman:
No, not yet. It's just looking more beautiful, as the step functions 
produce the same (absolute) EMA slope values.

Cheers,
Roman

jjxtra:
I checked in a version of the moving average calculator that uses decimal. So now there is the MovingAverageCalculator and MovingAverageCalculatorDecimal classes. I think we can close this pull request.

PR #58, opened 2018-03-13
Fix withdrawal bugs at Bitfinex and Bittrex. Standardize naming

bichuga:
Add GetWithdrawalFees for Bitfinex
Standardize naming across Exchange* models.
Add ToStrings to models


jjxtra:
Thanks for doing this. Having withdraws is very handy!

PR #60, opened 2018-03-14
Implemented private methods of Bitstamp REST API

TarasMi:


jjxtra:
Looks like some of a merge is still in the first file

jjxtra:
Thanks so much for putting this in!

PR #66, opened 2018-03-16
Master

vdboots:
added support for xrp withdraw


vdboots:
this is only for bitstamp

PR #77, opened 2018-03-27
Introduce RoundDown helper to assist in buying max coins when precision is not available

bichuga:
Create the beginnings of a unit test project


jjxtra:
Thanks for putting this together. Probably about time we had a unit test project! :)

PR #79, opened 2018-03-28
Nonce generation

espirulina:
Multiply by 10 - Fixes different Time-zones issue

### DIFF
--- a/ExchangeSharp/API/BaseAPI.cs
+++ b/ExchangeSharp/API/BaseAPI.cs
@@ -208,33 +208,32 @@ namespace ExchangeSharp
                     switch (NonceStyle)
                     {
                         case NonceStyle.Ticks:
-                            nonce = now.Ticks;
+                            nonce = now.Ticks * 10;
                             break;
 
                         case NonceStyle.TicksString:
-                            nonce = now.Ticks.ToStringInvariant();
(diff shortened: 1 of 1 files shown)


jjxtra:
The nonce uses DateTime.UtcNow and *should* be unaffected by time zones. I'm worried that this is simply a hack to work around the underlying problem, which may re-appear. Is it possible to remove the * 10 and record your screen and debug through the code and post a video of the problem?

espirulina:
image

jjxtra:
The first number has 19 digits, the second 18. Does this error happen if
you re-generate new API keys?

-- Jeff

espirulina:
@jjxtra Thanks ! Is the above picture enough, or you really need a video ?

jjxtra:
Picture is fine. Does it work if you generate new API keys?

-- Jeff

espirulina:
thanks @jjxtra ! It worked ! My guess is that somehow Keys generated before Summer time change (last sunday) might lead to wrong generation of bounces after that event. Does this make sense in your opinion ? :/ Take care !

jjxtra:
I'm not sure this is the problem. I have API keys from several months ago
on Poloniex and they work fine.

-- Jeff

PR #81, opened 2018-03-29
Added clamping of Binance quantity and prices

sthewissen:
Binance has limitations on the amount of decimals provided for quantity and price. Each coin has its own settings of how many decimals both can have which are defined in the market metadata. This adds clamping to the quantity and prices if a quantity or price with more decimals than accepted by Binance is provided.

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -383,8 +383,17 @@ namespace ExchangeSharp
             outputPrice = order.Price;
 
             // Get the exchange markets if we haven't gotten them yet.
+
             if (_exchangeMarkets == null)
-                _exchangeMarkets = GetSymbolsMetadata();
+            {
+                lock (this)
+                {
+                    if (_exchangeMarkets == null)
+                    {
+                        _exchangeMarkets = GetSymbolsMetadata();
+                    }
+                }
+            }
 
             // Check if the current market is in our definitions.
             ExchangeMarket market = _exchangeMarkets.FirstOrDefault(x => x.MarketName == symbol);
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -139,8 +139,8 @@ namespace ExchangeSharp
                 JToken priceFilter = filters?.FirstOrDefault(x => string.Equals(x["filterType"].ToStringUpperInvariant(), "PRICE_FILTER"));
                 if (priceFilter != null)
                 {
-                    market.MaxPrice = lotSizeFilter["maxPrice"].ConvertInvariant<decimal>();
-                    market.MinPrice = lotSizeFilter["minPrice"].ConvertInvariant<decimal>();
+                    market.MaxPrice = priceFilter["maxPrice"].ConvertInvariant<decimal>();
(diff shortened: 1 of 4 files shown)


ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
```diff
@@ -367,16 +372,39 @@ public override IEnumerable<MarketCandle> GetCandles(string symbol, int periodSe
 
         public override ExchangeOrderResult PlaceOrder(ExchangeOrderRequest order)
```
bichuga: Would it be worth also catching the specific errors thrown by the Binance webservice (LOT_SIZE/PRICE_FILTER) and reissuing the request?

ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
```diff
@@ -367,16 +372,39 @@ public override IEnumerable<MarketCandle> GetCandles(string symbol, int periodSe
...
         {
+            decimal outputQuantity, outputPrice;
             string symbol = NormalizeSymbol(order.Symbol);
             Dictionary<string, object> payload = GetNoncePayload();
             payload["symbol"] = symbol;
             payload["side"] = (order.IsBuy ? "BUY" : "SELL");
             payload["type"] = order.OrderType.ToString().ToUpperInvariant();
-            payload["quantity"] = order.RoundAmount();
+
+            outputQuantity = order.RoundAmount();
```
bichuga: Should respect the "ShouldRound" parameter

ExchangeSharp/Utility/CryptoUtility.cs
```diff
@@ -149,6 +149,35 @@ public static SecureString StringToSecureString(string unsecure)
             return secure;
         }
 
+        public static decimal ClampQuantity(decimal minQuantity, decimal maxQuantity, decimal? stepSize, decimal quantity)
```
bichuga: maybe do some more validation of these inputs?

ExchangeSharp/Utility/CryptoUtility.cs
```diff
@@ -149,6 +149,35 @@ public static SecureString StringToSecureString(string unsecure)
...
+            {
+                quantity = Math.Min(maxQuantity, quantity);
+                quantity = Math.Max(minQuantity, quantity);
+                quantity -= quantity % stepSize.Value;
+                quantity = Floor(quantity);
+            }
+            return quantity;
+        }
+
+        public static decimal ClampPrice(decimal minPrice, decimal maxPrice, decimal? tickSize, decimal price)
```
bichuga: How about adding unit tests for these methods to prove that they're working exactly as you think and prevent others from breaking them?

bichuga:
To make sure I understand this change, this is all based off of "StepSize" right? If someone sends in 1.2345678 and the step size is 0.001, the number will be changed to 1.234? If the stepsize is 0.005, it will become 1.23?

Is it worth doing min/maxprice clamping if stepsize isn't provided?

sthewissen:
Yes it's based on stepsize for quantity and ticksize for price. Your sample is correct, because if you would give 1.2345678 into a Binance order it will return an errorcode. Since stepsize and ticksize were already in the project as nullable decimals I was hesitant to change those. There is probably a good reason they're nullable.

jjxtra:
Thanks for the back and forth on this. Let me know when you feel it is ready to merge and I will have a look over.

sthewissen:
I wrapped it in a lock as described 👍

jjxtra:
I changed the clamp method to not call round down, it produced invalid results given the step size provided. They all pass now.

PR #90, opened 2018-04-07
Polo refactoring (+ fix rounding)

Ofekw:
ExchangePoloniexAPI is getting too complicated to edit safely. Begin refactoring to bring it under test.

Pulled out MakeRequest into a RequestHelper to make it easier to mock during unit testing. The helper is now initialized during the api constructor or passed in.

Also refactored parse order details in Polo. Lots of duplicate code removed. Added code coverage for everything that was refactored.

### DIFF
--- a/ExchangeSharp/API/APIException.cs
+++ b/ExchangeSharp/API/APIException.cs
@@ -1,0 +1,23 @@
+﻿namespace ExchangeSharp
+{
+    using System;
+
+    /// <summary>
+    /// Exception class for API calls
+    /// </summary>
+    public class APIException : Exception
+    {
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="message">Message</param>
+        public APIException(string message) : base(message) { }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="innerException">Inner exception</param>
+        public APIException(string message, Exception innerException) : base(message, innerException) { }
+    }
+}
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -22,8 +22,6 @@ using Newtonsoft.Json.Linq;
 
 namespace ExchangeSharp
 {
-    using ExchangeSharp.API.Services;
-
     public class ExchangeBinanceAPI : ExchangeAPI
     {
         public override string BaseUrl { get; set; } = "https://api.binance.com/api/v1";
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -778,7 +778,7 @@ namespace ExchangeSharp
             }
         }
 
-        protected override void ProcessRequest(HttpWebRequest request, Dictionary<string, object> payload)
+        public override void ProcessRequest(HttpWebRequest request, Dictionary<string, object> payload)
         {
             if (CanMakeAuthenticatedRequest(payload))
             {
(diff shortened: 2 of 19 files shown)


ExchangeSharp/API/IRequestHelper.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ExchangeSharp.API.Services
```
bichuga: Change namespace to remove .Services

ExchangeSharp/API/APIException.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ExchangeSharp
```
bichuga: don't forget header

ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
```diff
@@ -22,6 +22,8 @@ MIT LICENSE
 
 namespace ExchangeSharp
 {
+    using ExchangeSharp.API.Services;
```
bichuga: Remove all of these excess usings after namespace is fixed

ExchangeSharp/API/Exchanges/ExchangePoloniexAPI.cs
```diff
@@ -302,13 +308,16 @@ public override IEnumerable<ExchangeMarket> GetSymbolsMetadata()
...
+        }, ...
              */
 
             var markets = new List<ExchangeMarket>();
             Dictionary<string, JToken> lookup = MakeJsonRequest<Dictionary<string, JToken>>("/public?command=returnOrderBook&currencyPair=all&depth=0");
             foreach (var kvp in lookup)
             {
+                // StepSize is 8 decimal places for both price and amount on everything at Polo
+                const decimal StepSize = 0.00000001m;
+                const decimal minTradeSize = 0.0001m;
```
bichuga: Move outside the foreach

ExchangeSharp/API/RequestHelper.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ExchangeSharp.API.Services
+{
+    using System.IO;
+    using System.Net;
+
+    class RequestHelper : IRequestHelper
```
bichuga: public

bichuga: Add documentation

jjxtra:
Can we make the RequestHelper a public get; set; property and have it default to the http implementation? If set to null it should choose the default http implementation as well.

This way constructor overloads are not necessary, and a mocking framework can freely change it.

I'd also like to put any dependancy methods that the request helper is using behind an interface instead, that way they aren't publicly exposed to anyone using an instance of an API class. Having them protected served this purpose, but now they are dependant methods of another class, we need an interface. We will need to define an IRequestHelper interface, or some better name if you think of one :)

Example: Uri IRequestHelper.ProcessRequestUrl(UriBuilder url, Dictionary<string, object> payload);

The method only shows up if the instance is of type IRequestHelper.

Let me know your thoughts, counter-arguments, changes, etc.

bichuga:
@jjxtra You have it right that all of the request helper stuff needs to move out of BaseAPI. The plan is to put them in IRequestHelper but once we do we need to also move each API's ProcessReq and ProcessResp logic. These would probably be implemented as subclasses of RequestHelper. This refactor is already enormous just to get Polo testable so we decided to cap it here. Get this in so people can start unit testing their contributions.

I don't have a strong opinion of using a property or a constructor parameter for setting the IRequestHelper.  (I'm not even internally consistent on this). Would be happy to switch that if you prefer.

Ofekw:
@jjxtra I'll have a think about what you said and have a code update on here tomorrow no worries. Cheers for the feedback.

jjxtra:
Let's switch to the property. We can leave the methods public and do a follow up pull request.

jjxtra:
Merging this pull request. I'll do a follow up checkin.

jjxtra:
I've pushed my changes. Request helper is now APIRequestMaker and I also added IAPIRequestHandler interface to only give the APIRequestMaker class the minimum it needs to make a request. Let me know any thoughts or ideas for improvement.

jjxtra:
I think we are good on this refactor for now. I like the protected methods with override. May revisit layer but unit tests are easy now.

PR #96, opened 2018-04-18
Use async methods by default, sync wraps async

jjxtra:
Before this change, async calls were simply wrapped in Task.Run calls. Now async calls are the base implementation, and the synchronous calls wrap the async calls. This will avoid deadlock/thread starvation. Any future api methods should be implemented with the same pattern.


jjxtra:
Going to leave this pull request open until the afternoon to get eyes on it. Will merge around 5 pacitic time (US).

kamelmansouri:
The WebSocket protocol is not supported on this platform???
i used .net 4.7

PR #99, opened 2018-04-21
Added Abucoins Exchange

BobDeCuir:
Implemented OnGetHistoricalTradesAsync with new pattern. On other ExchangeSharp exchanges  this call should always check if the callback is null before sending results because the calling object may be disposed by the time the call is complete.

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
@@ -616,6 +616,11 @@ namespace ExchangeSharp
         public const string Gemini = "Gemini";
 
         /// <summary>
+        /// Gemini
+        /// </summary>
+        public const string Hitbtc = "Hitbtc";
+
+        /// <summary>
         /// Kraken
         /// </summary>
         public const string Kraken = "Kraken";
--- a/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
@@ -616,7 +616,7 @@ namespace ExchangeSharp
         public const string Gemini = "Gemini";
 
         /// <summary>
-        /// Gemini
+        /// Hitbtc
         /// </summary>
         public const string Hitbtc = "Hitbtc";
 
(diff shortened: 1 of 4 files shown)


BobDeCuir:
Added BleuTrade exchange. Please note the OnGetCandlesAsync. Unlike other ExchangeSharp exchanges, it doesn't throw exceptions if the periodSeconds or limit parms don't match what the exchange expects - it reassigns them accordingly. This way calls to multiple exchanges can use the same parms without error.

BobDeCuir:
I'll try to be more careful with the copy/paste. I'm working with three open solutions to get these changes updated and it gets confusing. 

"Should the timestamp parameter really be an (int) cast of seconds? This would only allow one API call per second? Or does this API allow duplicate timestamp values as long as they are greater than or equal?"

I'm not sure which Exchange and which call you're referring to.

jjxtra:
ExchangeAbucoinsAPI.cs line 32 or so.

BobDeCuir:
If I remember correctly, I put that in to remove the decimal places on the UnixTimeStamp.

jjxtra:
Looks like it's good, they say it has to be +/- 30 seconds from current
time to be valid, no mention of uniqueness.

-- Jeff

BobDeCuir:
Added HitBtc Exchange. See comments in code regarding open orders vs historical trades' OrderIds and Bank vs Trading accounts.

BobDeCuir:
Kewl. More to come. Next Batch: Cryptopia, KuCoin, LiveCoin, TuxExchange, Yobit

jjxtra:
Awesome! I did a cleanup / refactor on the first pull request. You can check my changes to get an idea from latest master commit. I found some problems in GetHistorical trades and ticker calls that were not getting the volume. Basically if you can run the console app with the test parameter and everything works, then you are probably OK at least on your public end points.

You can run test exchangeName=ExchangeHitbtc for example to just test one exchange instead of all of them. You'll need to pull latest master to get the exchangeName parameter.

Thanks again for sharing your code!

BobDeCuir:
Thank You! Always good to have more eyes on the code. I have my own unit tests for the APIs I use in my own software, but I don't use all of the ExchangeAPIs, so a few functions that I added to the fork weren't tested enough. I'll run your tests for the remainder of Exchanges I add (as well as my own).

PR #105, opened 2018-04-25
optimize ExchangeKrakenAPI.OnGetTickersAsync()

riversx:
This optimization is made to get all tickers in a single call to the kraken API.


jjxtra:
Very nice!

PR #107, opened 2018-04-25
Added Kucoin Exchange

BobDeCuir:
Notes: The GetCandles API is a PRIVATE call on Kucoin and requires an API Key and API Secret
The Cancel Order function has not been tested in production
See comments in code regarding OrderIds and Orginal OrderIds
See additional comments in code


BobDeCuir:
That was fast! I caught an error right after I created the pull request. Line #419 should read "coinTypePair" not "CoinTypePair". Can you correct this, or do I need another upload?

jjxtra:
I am fixing it

-- Jeff

BobDeCuir:
Thanks, Jeff.

PR #110, opened 2018-04-27
Added Yobit Exchange

BobDeCuir:
Added Yobit Exchange
See comments in code

Fixed unix timestamps not returning time portion of DateTime

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
@@ -644,5 +644,10 @@ namespace ExchangeSharp
         /// Poloniex
         /// </summary>
         public const string Poloniex = "Poloniex";
+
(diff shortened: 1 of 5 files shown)


jjxtra:
Thanks again for this. I hope your trading is going well :)

BobDeCuir:
You're very welcome, and thank you for the continued maintenance. 

I've got one last Exchange to add (LiveCoin) and that will be it for a while. I'll add more later. I'm only interested in Exchanges with trading APIs AND that allow deposits/withdraws via API. There's a handful more, but I have to get back to my trading software.

Markets are up. Trading is good. Let's all retire wealthy ;)

jjxtra:
Did a little cleanup on the last pull request but all looking good. This will come in handy for multiple apps and websites I am planning. Thanks again for sharing.

Good luck! :)

PR #115, opened 2018-04-29
Add paramter symbol for CancelOrder and GetOrderDetails

wukan1986:


jjxtra:
Thanks for doing this!

PR #118, opened 2018-05-01
Add Huobi exchange

wukan1986:


jjxtra:
All looks good except one thing- use StringComparer.Ordinal, no need to make your own StringComparer class. Unless I am missing something...

wukan1986:
the SortedSet sort the key IgnoreCase.
so it will get keys like this.   A a B  b C c
but the exchanges huobi and okex must use keys like this  A B C a b c

PR #124, opened 2018-05-05
Poloniex Symbol is not reversed

LORDofDOOM:
add Symbol to KuCoin to allow GlobalSymbolToExchangeSymbol


jjxtra:
Do the unit tests pass with these changes?

PR #135, opened 2018-05-17
Add IExchangeAPI.GetOrderBookWebSocket

superhappychris:
Added a new IExchangeAPI method for streaming the order book.  This has been implemented for both Poloniex and Binance.  I also created a test in the ExchangeSharpConsole project.


jjxtra:
Awesome, thanks for sharing!

PR #140, opened 2018-05-19
Change ConnectWebSocket messageCallback

wukan1986:


jjxtra:
Thanks for the pull request. I checked in a change to make web socket raw bytes. I added convenience extension method UTF8String() for byte[] and each exchange now uses that. For your case, you can decompress the bytes then use the same convenience method.

PR #149, opened 2018-05-24
Support arbitrary new currencies for deposit/withdrawal at Bitfinex

bichuga:
I spoke with Bitfinex support and they agreed that going forward they weren't going to do special expansion of currency symbols to currency names for the deposit method. It would just be the currency symbol in lowercase. Old lookups like BTC -> bitcoin would still apply and weren't going to be backfilled with BTC -> btc.


jjxtra:
Thanks for this. What a horrible choice on their part to do this in the first place

PR #150, opened 2018-05-24
Fix for https://github.com/jjxtra/ExchangeSharp/issues/120 

bichuga:
Need to force a refresh of cached GetSymbolsMetadata if no result is found during ClampOrderPrice/Quantity

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
@@ -31,9 +31,10 @@ namespace ExchangeSharp
         /// </summary>
         public const char GlobalSymbolSeparator = '-';
 
+        protected IEnumerable<ExchangeMarket> exchangeMarkets;
+
         private static readonly Dictionary<string, IExchangeAPI> apis = new Dictionary<string, IExchangeAPI>(StringComparer.OrdinalIgnoreCase);
 
-        private IEnumerable<ExchangeMarket> exchangeMarkets;
         private bool disposed;
 
         /// <summary>
@@ -43,26 +44,38 @@ namespace ExchangeSharp
         /// <returns>The ExchangeMarket or null if it doesn't exist</returns>
         protected ExchangeMarket GetExchangeMarket(string symbol)
         {
-            PopulateExchangeMarkets();
-            return exchangeMarkets.FirstOrDefault(x => x.MarketName.EqualsWithOption(symbol));
+            bool refreshed = PopulateExchangeMarkets(false);
+            ExchangeMarket market = exchangeMarkets.FirstOrDefault(x => x.MarketName.EqualsWithOption(symbol));
+            if (market == null && !refreshed)
+            {
+                PopulateExchangeMarkets(true);
+                market = exchangeMarkets.FirstOrDefault(x => x.MarketName.EqualsWithOption(symbol));
+            }
+
+            return market;
         }
 
         /// <summary>
         /// Call GetSymbolsMetadata and store the results.
         /// </summary>
-        private void PopulateExchangeMarkets()
+        private bool PopulateExchangeMarkets(bool forceRefresh)
         {
+            bool refreshed = false;
+
             // Get the exchange markets if we haven't gotten them yet.
-            if (exchangeMarkets == null)
+            if (forceRefresh || exchangeMarkets == null)
             {
                 lock (this)
                 {
-                    if (exchangeMarkets == null)
+                    if (forceRefresh || exchangeMarkets == null)
                     {
                         exchangeMarkets = GetSymbolsMetadata();
+                        refreshed = true;
                     }
                 }
             }
+
+            return refreshed;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


jjxtra:
Looks pretty good. A couple suggestions:

- make exchange markets variable private
- make exchange markets variable a dictionary<string, exchangemarket>
- make protected method to add and remove exchange markets

Let me know if that makes sense or if I am missing something in your pattern.

jjxtra:
Maybe the add/remove method isn't needed as the populate exchange markets function takes care of it. The dictionary is just to optimize the GetExchangeMarket not having to loop all the markets.

bichuga:
Dictionary definitely makes sense here. 

I don't think an add/remove method for exchangemarkets makes sense because callers (who aren't unit tests) should have no reason to use that. 

I made exchange markets variable protected to enable unit testing and I'm hesitant to switch it back to private because that would really complicate the unit tests. How strongly do you feel about this? Can you think of a better way to have the best of both worlds?

jjxtra:
```
public IReadOnlyDictionary<string, ExchangeMarket> ExchangeMarkets { get; private set; }
```
I think this makes the most sense. Unit tests and other future functions can read from and validate the data in the dictionary. GetSymbolsMetadata doesn't need to be changed, we can just iterate the result and build up a new dictionary in ```PopulateExchangeMarkets``` and then assign it to ```ExchangeMarkets ```

For the unit test to be most valuable, it needs to trigger the normal flow of calling PopulateExchangeMarkets, which happens in the PlaceOrderAsync API calls when the clampXXX methods are called.

This would entail mocking the response from PlaceOrderAsync, along with mocking the response from OnGetSymbolsMetadataAsync. So the JSON from both would need to be captured from the live system and then mocked in. The Poloniex unit tests show how this is done.

The unit tests would call place order, validate that a good order came back with the right price and quantity and validate that ExchangeMarkets has the expected markets in them.

As it stands now, the unit test is just stuffing a few markets into the protected variable which is not how things will work when using the PlaceOrderAsync call against the live exchanges.

I realize this is a little bit more work, and I'm happy to assist where needed, let me know what you think.

Thanks for getting the effort started.

bichuga:
What you're proposing is definitely possible, and it is also definitely not a unit test :). The behavior I'm ensuring is the very targeted caching and refreshing of markets when using GetExchangeMarket. The tests are ~5 lines long and very clear on what behavior is intended. (Incidentally it's also clear if I've missed scenarios or not considered something.) The fact that we have to make uncomfortable decisions here about what internal functionality to expose is a side effect of not having built this library using TDD from the start.

Like you say, I could instead get at this by calling PlaceOrder, allowing Clamp to run, mocking the response from OnGetSymbolsMetadata and finally validating ExchangeMarkets. This has the side effect of tying all these components together so the next dev can't change PlaceOrder without potentially having to rewrite the GetExchangeMarket caching logic tests. This is obviously not what we want. 

In an ideal world, we'd be using Single Responsibility Principle and we'd have two sets of unit tests: One set that validates the internals of an IExchangeMarketFetcher; and a separate set that validates that IExchangeMarketFetcher is called at appropriate times. (In my professional life this is the approach I would have taken.)

jjxtra:
Good points @bichuga. Thanks for discussing. Let's worry about just testing the public interface for now.  So that would mean just testing ```GetExchangeMarketFromCacheAsync```, which I have promoted to a public method in the latest commit, along with proper async handling. Seems like this could be a useful function outside of order processing. I haven't made it part of ```IExchangeAPI``` yet, do you think that is necessary?

This does eliminate the need for a ```MockExchangeAPI```  class. It does however require that we provide mock data for the symbols metadata end point. I will use your unit test as a guide and implement the unit tests. For now, the current code solves the caching problem.


Thanks for the pull request!

jjxtra:
@bichuga I checked in code for the poloniex unit tests that add a test around the new method. Let me know if you have any feedback or suggestions for improvement.

Snippet:

```
        [TestMethod]
        public void GetExchangeMarketFromCache_SymbolsMetadataCacheRefreshesWhenSymbolNotFound()
        {
            var polo = CreatePoloniexAPI();

            // retrieve without BTC_BCH in the result
            polo.RequestMaker.MakeRequestAsync(null).ReturnsForAnyArgs(Resources.PoloniexGetSymbolsMetadata1);
            polo.GetExchangeMarketFromCache("XMR_LTC").Should().NotBeNull();
            polo.GetExchangeMarketFromCache("BTC_BCH").Should().BeNull();

            // now many moons later we request BTC_BCH, which wasn't in the first request but is in the latest exchange result
            polo.RequestMaker.MakeRequestAsync(null).ReturnsForAnyArgs(Resources.PoloniexGetSymbolsMetadata2);
            polo.GetExchangeMarketFromCache("BTC_BCH").Should().NotBeNull();

            // and lets make sure it doesn't return something for null and garbage symbols
            polo.GetExchangeMarketFromCache(null).Should().BeNull();
            polo.GetExchangeMarketFromCache(string.Empty).Should().BeNull();
            polo.GetExchangeMarketFromCache("324235!@^%Q@#%^").Should().BeNull();
            polo.GetExchangeMarketFromCache("NOCOIN_NORESULT").Should().BeNull();
        }
```

Issue #153, opened 2018-05-26
WriteToRequest UTF8 encoding confuses Polo signature validation


bichuga:
Commit 53fc836de9c5c597e0efd4957605af21e44309b2 change of encoding in BaseAPI.WriteToRequest from ASCII to UTF8 breaks Poloniex signing. (No idea why)

Was this change to fix a different issue?

jjxtra:
Checked in a change with utf-8 without prefix bytes and verified Poloniex private API. Let me know if you see any issues.

PR #154, opened 2018-05-26
Change encoding back to ASCII to fix Poloniex signature validation

bichuga:
#153


jjxtra:
Probably the stupid .net utf-8 prefix chars, I will do a proper utf-8 without later.

PR #156, opened 2018-05-28
Add margin api

superhappychris:
Added a new interface IMarginExchangeAPI for use by exchanges that allow margin accounts and trades.  Any exchange class that wants to take advantage of this new API should change their base class from ExchangeAPI to MarginExchangeAPI (which itself inherits from ExchangeAPI).  I included a few methods to start with and implemented these methods in the ExchangePoloniexAPI class.  Since these all require the private api, I didn't include any tests in the test console; however, I did test them independently with a live Poloniex margin account.

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeAPI.cs
@@ -444,7 +444,7 @@ namespace ExchangeSharp
         /// </summary>
         static ExchangeAPI()
         {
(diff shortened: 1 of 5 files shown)


jjxtra:
I've seen other projects that like to split different groups of functionality into different sub-classes and interfaces. I have chosen not to go that route in ExchangeSharp. How would you feel about just putting everything under IExchangeAPI and ExchangeAPI? We can group the margin calls with #region in ExchangeAPI.cs. Let me know if you have strong feelings either way. Thanks so much for doing this pull request.

PR #157, opened 2018-05-28
Add margin api take 2

superhappychris:
Removed IMarginExchangeAPI and placed all new logic inside IExchangeAPI per comments on the original pull request


jjxtra:
Looks like there are some merge conflicts. I put all the code in ExchangeAPI.cs and IExchangeAPI.cs into regions, that's probably why.

Issue #170, opened 2018-06-03
Binance does not accept quantities with more than 20 decimal places


jazzonaut:
I have a [BUG REPORT]

Details:
Binance does not accept quantities with more than 20 decimal places when placing an order.

PR #171, opened 2018-06-03
#170 - Binance does not accept quantities with more than 20 decimal p…

jazzonaut:
Binance does not accept quantities with more than 20 decimal places #170

### DIFF
--- a/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
@@ -533,7 +533,8 @@ namespace ExchangeSharp
             decimal outputQuantity = await ClampOrderQuantity(symbol, order.Amount);
             decimal outputPrice = await ClampOrderPrice(symbol, order.Price);
 
-            payload["quantity"] = outputQuantity;
+            // Binance does not accept quantities with more than 20 decimal places.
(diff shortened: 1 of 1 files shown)


ExchangeSharp/API/Exchanges/ExchangeBinanceAPI.cs
```diff
@@ -533,7 +533,8 @@ protected override IDisposable OnGetOrderBookDeltaSocket(Action<ExchangeSequence
             decimal outputQuantity = await ClampOrderQuantity(symbol, order.Amount);
             decimal outputPrice = await ClampOrderPrice(symbol, order.Price);
 
-            payload["quantity"] = outputQuantity;
+            // Binance does not accept quantities with more than 20 decimal places.
+            payload["quantity"] = Math.Round(outputQuantity, 20);
```
bichuga: Why does ClampOrderQuantity fail in this case?

jazzonaut: ClampOrderQuantity doesn't consider the number of decimal places.
The symbols metadata does come with LOT_SIZE filter, but because it is treated as decimal, that information is lost. Perhaps there needs to be a new variable MaxDecimalPlaces that will be extracted from LOT_SIZE.

PR #173, opened 2018-06-03
#172 - Binance - No support for getting current account's trades

jazzonaut:
Binance - No support for getting current account's trades #172


jjxtra:
Can you explain your use case for getting individual trades vs. the orders which aggregate the trades?

jjxtra:
Since only binance supports this can we delete the changes from iexchangeapi and move the changes from exchangeapi into exchangebinanceapi?

PR #184, opened 2018-06-20
Bug in new order

Peixer:
amount need have decimal and price need value


jjxtra:
Thanks for the pull request, I have already put the changes in from your email :)

PR #194, opened 2018-06-26
Fixed WebSocketWrapper not disposing properly

jazzonaut:


jjxtra:
Thanks for taking the time!

PR #196, opened 2018-06-28
Implement order book delta socket for Bittrex

bichuga:
Since this is a very high traffic websocket, I wanted real objects to parse the JSON into. I borrowed these from [bittrex.net](https://github.com/JKorf/Bittrex.Net).

A couple notes: I don't love that these model object classes are all in one giant file. It was also confusing to navigate the ExchangeBittrexApi.cs file that has the exchange implementation classes as well as the various websocket classes. Please accept my vote to split each of these classes out into their own files or reorganize this in some way. 

Related: I approve of the refactoring you did around GetExchangeOrderBookDeltas. Including the symbol and sequence number in the order book itself makes things way cleaner. 👍


jjxtra:
I re-organized the files and did some code cleanup. Each exchange is in their own folder, with model objects and helper classes underneath a partial class for the exchange class. Let me know what you think.

bichuga:
Looking way better. Thanks for doing the refactor.

I don't have a strong opinion, just curious because I haven't seen it this
way before... Why use nested classes under an exchange rather than just
putting the related objects in their own namespace?

jjxtra:
I suppose that could work as well and make them internal. This way they don't get in the way, and I've been hesitant to do extra namespaces because I always find it confusing. But either way would work fine :)

PR #213, opened 2018-07-20
Rebrand GDAX -> Coinbase and Implement OnGetOrderBookDeltasWebSocket

bichuga:
Coinbase is killing their GDAX api endpoints at the end of this year. Make the switch now to their new branding to avoid being broken.

Implemented OrderBook websocket for GDAX

Other minor changes:
*   Added console.out tracing to signalR sockets
*   Added a LastUpdatedUTC to order books


jjxtra:
Wow nice work. One request, let's hide that namespace so that nobody outside the ExchangeSharp.dll can see it.

namespace ExchangeSharp
{
    namespace ExchangeSharp.Coinbase
    {
    }
}

And make all the classes internal ... instead of public ...

Any of the exchange api methods should only be using public common models so I *think* this should work.

Is there a chance we'd want to re-use any of the Coinbase.* objects / classes you made?

bichuga:
What you propose doesn't actually hide the namespace. Someone from outside can just do a 
`using ExchangeSharp.ExchangeSharp.Coinbase` (per your example above). internal classes are the right answer for hiding these objects.

The coinbase model objects are specific to the json from their APIs and I think it very unlikely that we'd want to reuse them elsewhere.

PR #215, opened 2018-07-23
Lock the order book before applying deltas so consumers can lock if desired to prevent concurrency issues

bichuga:


jjxtra:
Merged manually.

PR #217, opened 2018-07-26
Changed volume and pair to actual BTC volume and coin name

sthewissen:
The ticker endpoint returns the quote volume and quote currency but these weren't being set for KuCoin. Instead it just sets the base currency and volume twice.


jjxtra:
Thanks for the fix!

sthewissen:
No problem :) Looking forward to 0.5.4 :)

PR #225, opened 2018-08-01
Implement Binance GetCurrencies and separate ExchangeCurrency IsEnabled into DepositEnabled and WithdrawalEnabled

bichuga:
**Breaking Change**: Separate ExchangeCurrency IsEnabled into DepositEnabled and WithdrawalEnabled since multiple exchanges provide this granularity.

* Implement Binance GetCurrencies using undocumented API advertised in Binance Telegram channel.
* Refactor Binance models into separate files and namespace.


jjxtra:
Can you make those Binance model classes internal?

bichuga:
I didn't initially because it breaks the tests, but on the other hand the
tests are some silly deserialization tests I wrote during development which
are almost not worth preserving. Unfortunately we've got no other coverage
of that code so I settled on Internal+InternalsVisibleTo ExchangeSharpTests

jjxtra:
Thanks for taking the time.

PR #229, opened 2018-08-04
Fix json deserialization bugs in Binance GetCurrencies API

bichuga:
Refactoring and testability improvements for ExchangeAPI.


jjxtra:
Why are the synchronous methods virtual now? They should always forward the call to the async virtual method...?

jjxtra:
This may be a code smell. Should we nuke the synchronous implementations completely and make an extension method to make calling an async method synchronously easier? Something that does the GetAwaiter().GetResult() in a single call?

Ideas...

```csharp
api.GetTickersAsync().Sync();
api.GetTickersAsync().Synchronously();
api.GetTickersAsync().Wait();
api.GetTickersAsync().Result();
```

Then there is no ambiguity, all you do is plop on the Sync extension and it waits. No multiple paths, less confusing.

Any counter?

ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
```diff
@@ -443,12 +235,12 @@ public void Dispose()
             {
                 disposed = true;
                 OnDispose();
-                Cache.Dispose();
+                Cache?.Dispose();
```
jjxtra: Cache should never be null. Do you have a case where this is null?

jjxtra:
I pulled in your file splitting out changes and a few others. You may want to re-sync from master.

ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
bichuga: This nullref'd on me. No idea why

jjxtra: That is very odd. Here it the definition:

```csharp
protected MemoryCache Cache { get; } = new MemoryCache();
```

bichuga:
The async extension methods might be cool. Something to bundle up the ConfigureAwait GetAwaiter business would definitely be nice. There's probably something clever that could be done here.

jjxtra:
I added a Sync extension and I enjoy using it very much, unless there is a compelling reason I say we delete the synchronous methods. Anyone else want to way in before we start chopping heads?

Ofekw:
chop heads!

ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
bichuga: Can it dispose twice?

jjxtra:
I'm cleaning out the synchronous methods and found about 40 API calls that were being done synchronously in async methods! I like the new Async extension method because it makes synchronous code stand out a lot more, and makes it much harder to accidently write synchronous code in an async method. Win win.

It also calls out code where the architecture could be better. Anything outside of a unit test or a simple console app demo should probably not be using synchronous code unless there is a specific edge case, otherwise it's a sign the code needs to be refactored.

jjxtra:
It's done, heads chopped.

ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
jjxtra: I suppose it's possible. The Dispose in MemoryCache is wrapped in a try/catch so *shouldn't* be a problem.

jjxtra:
@bichuga Can you remind me why the public methods in ExchangeAPI need to be virtual at this point?

bichuga:
When mocking an abstract class you can only override the behavior of
abstract or virtual methods

jjxtra:
So if we make ExchangeAPI non-abstract then we can make the methods non-virtual?

Can you mock it with a private constructor?

bichuga:
The other option is to put everything* on the interface and Substitute that
instead. The downside is the mock won't have the benefits of the default
behavior provided by the abstract class. Maybe not a big deal? It comes
down to what people want to test.

For example, maybe your test is you ask one mock exchange for their tickers
and they return STR for Stellar. You ask a different fake exchange and they
return XLM. You want to see if your program matches these two up and the
way it does that is by calling the ExchangeSymbolToGlobalSymbol method. If
your mock exchange is of the abstract class, you get some default behavior
which may match these for you. If you're mocking the interface, this method
will have no implementation and depending on your testing framework this
would return null or notimplemented.

It's a bit of a convoluted example because (a) that's a pretty lame unit
test and (b) the symbol normalization stuff ought to live in a different
class anyway, but that's the type of tradeoff you'd be making.

*This would include important things from BaseAPI that might be called by
the user (ex. RateGate settings or Nonce offsets)

jjxtra:
I removed the virtual. I think this project will work for mocking the non-virtual methods: https://github.com/tonerdo/pose

Let me know if it looks like it will work. I'd really rather those not be virtual, as it can confuse people and they might override those methods for new exchanges.

bichuga:
I understand you'd like to avoid confusing developers who are implementing
new exchanges. Unfortunately, for the purposes of enabling unit testing,
run-time shimming is emphatically not the way.

This is MSFT's official position on the shimming framework they created
(the same thing that pose is doing). For reasons discussed in the article,
we've outright banned its use at work.
http://www.peterprovost.org/blog/2012/04/25/visual-studio-11-fakes-part-2/

Is your concern not something that could be enforced at pull request time?
There are so many examples of well-implemented exchanges that people can
follow. If these methods are virtual you can enable developers to create
super clear unit tests for their own code using your framework.

jjxtra:
Since those public are just wrappers of the on... protected virtual can those protected virtual be mocked instead? Or does having them protected prevent that? We could go protected internal to let the test framework see them?

jjxtra:
All the protected methods are now protected internal virtual, let me know if that is not sufficient for your testing purposes.

bichuga:
Protected won't work because the test framework can't override it. Internal
would require associated "InternalsVisibleTo" attribute pointing to
JoeUser's unit test assembly. A different option is to make On* methods
public, though I think this is even worse because that could definitely
confuse the callers. The way things are set up here, I sincerely
think virtual's our best bet to give people a clean unit testing interface
without overexposing APIs. Just keep an eye out for foolishness at code
review time when people are implementing new exchanges.

bichuga:
Take this scenario as your anchor:

Start from a new project.
Create a class. That class has a single method which calls GetTickers on
any exchange and Console.WriteLine(theReturnedObject).
From a different class, how much effort does it take to make
theReturnedObject be a single BTC ticker?

jjxtra:
Ok it's checked in as virtual again.

PR #233, opened 2018-08-06
Update ExchangeKrakenAPI.cs

username77:
added new symbols to GlobalSymbolToExchangeSymbol
edited QueryOrdersAsync to return correct opened trades, before has returned 0 rows
added OnGetSymbolsMetadataAsync to get coin info for precision amount and price


jjxtra:
Why is OnGetAmountsAvailableToTradeAsync commented out?

username77:
remove comments. maybe was my mistake

PR #253, opened 2018-09-13
Bittrex: added Trade stream (websocket)

vslee:
Bittrex doesn't currently send out FillId on socket.bittrex.com, only beta.bittrex.com, but this will be ready when they start
https://github.com/Bittrex/beta/issues/2
https://github.com/Bittrex/bittrex.github.io/issues/3
You can always change the URL on the top of the file (ExchangeBittrexAPI_WebSocket.cs) to beta.bittrex.com to start getting FillIds now


jjxtra:
Thanks. Will switch over once out of beta.

jjxtra:
Just heads up made the models internal, we don't want exchange specific public models sneaking out of framework :)

PR #261, opened 2018-10-09
Make cache optional

superhappychris:
I noticed that you added some forced caching into some of the methods. I'm not against having a caching mechanism available, but making it mandatory doesn't seem like the way to go. It should be up to the consumers of the library to specify if they need the calls cached or not.

That being said, in this code change I left the caching enabled by default on the GetSymbols methods because I'm pretty sure they've been that way for a while and I didn't want to break anyone's code. The other cache calls, however, are new so I thought it was best to leave the cache disabled by default.

Also, just a suggestion, but if you're going to include caching with the library, it would be better if the durations were configurable somehow instead of hard coded like they are now.

### DIFF
--- a/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
+++ b/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
@@ -402,27 +402,33 @@ namespace ExchangeSharp
         /// <summary>
         /// Get exchange symbols
         /// </summary>
+        /// <param name="cacheResponse">Cache the response?</param>
         /// <returns>Array of symbols</returns>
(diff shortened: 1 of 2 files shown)


jjxtra:
Thanks for submitting. I checked in some code just now that I think will be more robust and allows caching any public method easily. If you don't want caching, simply clear MethodCachePolicy property. See most of the public property in ExchangeAPI.cs for a cache example.

superhappychris:
Thanks for the feedback.  While I think your new code is an improvement, I still think it should be opt-in instead of opt-out like you have it now.

jjxtra:
I added a static bool to ExchangeAPI called UseDefaultMethodCachePolicy. Default is true, but simply set to false and cache goes away for any created ExchangeAPI.

PR #268, opened 2018-10-16
Added OnGetTickersWebSocket implementation to ExchangeKucoinAPI

superhappychris:
Also added the ability to specify specific ticker symbols when using the tickers websocket.


jjxtra:
params string[] symbols will not be null if it is left off entirely, we should check that ```params == null || params.Length == 0```

superhappychris:
ah yes, good catch! I've just checked in a fix for that.

superhappychris:
Whoops! Didn't mean to close the pull request!

superhappychris:
@jjxtra this should be good to go now.  Is there something else you want me to change before you merge this in?

PR #269, opened 2018-10-17
Fix null ref exception when using tickers web socket on Bittrex

superhappychris:
This was caused by the [most recent commit](https://github.com/jjxtra/ExchangeSharp/commit/83370161aac858e8d74068f31a3e0287771a3ce7).


vslee:
I'm assuming you construct a ClientWebSocket and set the KeepAlive prior to calling Start(). Thus getting a NRE. Good catch.

Issue #267, opened 2018-10-16
[BUG REPORT] Fix Currency Property Naming in ExchangeMarket class


superhappychris:
Right now the ExchangeMarket class uses the property names MarketCurrency and BaseCurrency to represent what is usually called the "base currency" and "quote currency" respectively.  This is extremely confusing and prone to causing bugs.  Just doing a quick search I found that at least the ExchangeAbucoinsAPI and ExchangeKrakenAPI are assigning the incorrect value to these properties...

I think we should rename MarketCurrency to BaseCurrency, and rename BaseCurrency to QuoteCurrency.  I'm happy to make the fix, but I realize this could be a breaking change for some people using this library, which is why I'm mentioning it here for discussion first.  Thoughts?

jjxtra:
Yea this is not consistant. I think there is another part in the framework that has them reversed too. So we should search for all instance of BaseCurrency and make sure they are correct. I am fine with the change if you want to pull request. Until we get to v. 1.0. I am ok with breaking changes.

vslee:
I agree with renaming for consistency with the market. I was not aware of this issue until reading your post. Perhaps there can be a notice placed on the front page for others to see.

superhappychris:
Ok sounds good. I took a closer look at the code and found 5 classes that have inconsistent property names that need to be changed.  I've listed them below by class.  The left hand side is the current name, the right hand side is what they will be renamed to.  Let me know if these look wrong or if there are any I'm missing.  In the meantime, I'll get started making the changes.

- ExchangeMarket
  - BaseCurrency : _QuoteCurrency_
  - MarketCurrency : _BaseCurrency_
- ExchangeTicker
  - BaseSymbol : _QuoteCurrency_
  - BaseVolume : _QuoteVolume_
  - ConvertedSymbol : _BaseCurrency_
  - ConvertedVolume : _BaseVolume_
- ExchangeTransaction
  - Symbol : _Currency_
- ExchangeDepositDetails
  - Symbol : _Currency_
- MarketCandle
  - BaseVolume : _QuoteVolume_
  - ConvertedVolume : _BaseVolume_

bichuga:
Along similar lines, my recollection is there is some inconsistency between
what we call a currrency "BTC" vs a market "FOO/BTC". Sometimes it's
referred to as a Ticker. Other times a Symbol. It would be really nice to
standardize this if it hasn't been yet.

superhappychris:
[...]

PR #272, opened 2018-10-19
Fix inconsistent naming

superhappychris:
This is a pretty large refactor attempting to bring some more consistent naming to several of the model class properties.  More specifically, every time a symbol is referred to in code it should be in reference to a currency pair, consisting of the following elements (order matters):  a base currency, an optional separator, and a quote currency.  For example, consider the symbol "BTC-USD"..  If BTC-USD==20,000.00; then 1 BTC is exchanged for 20000 USD. See issue #267 for more info.

### DIFF
--- a/ExchangeSharp/API/Exchanges/Abucoins/ExchangeAbucoinsAPI.cs
+++ b/ExchangeSharp/API/Exchanges/Abucoins/ExchangeAbucoinsAPI.cs
@@ -67,7 +67,7 @@ namespace ExchangeSharp
             throw new NotSupportedException("Abucoins does not provide data about its currencies via the API");
         }
 
-        protected override async Task<IEnumerable<string>> OnGetSymbolsAsync()
+        protected override async Task<IEnumerable<string>> OnGetMarketSymbolsAsync()
         {
             List<string> symbols = new List<string>();
             JToken obj = await MakeJsonRequestAsync<JToken>("/products");
@@ -78,7 +78,7 @@ namespace ExchangeSharp
             return symbols;
         }
 
-        protected override async Task<IEnumerable<ExchangeMarket>> OnGetSymbolsMetadataAsync()
+        protected override async Task<IEnumerable<ExchangeMarket>> OnGetMarketSymbolsMetadataAsync()
         {
             List<ExchangeMarket> markets = new List<ExchangeMarket>();
             JToken obj = await MakeJsonRequestAsync<JToken>("/products");
@@ -86,7 +86,7 @@ namespace ExchangeSharp
             {
                 markets.Add(new ExchangeMarket
                 {
-                    MarketName = token["id"].ToStringInvariant(),
+                    MarketSymbol = token["id"].ToStringInvariant(),
                     BaseCurrency = token["base_currency"].ToStringInvariant(),
                     QuoteCurrency = token["quote_currency"].ToStringInvariant(),
                     MinTradeSize = token["base_min_size"].ConvertInvariant<decimal>(),
@@ -99,10 +99,10 @@ namespace ExchangeSharp
             return markets;
         }
 
-        protected override async Task<ExchangeTicker> OnGetTickerAsync(string symbol)
+        protected override async Task<ExchangeTicker> OnGetTickerAsync(string marketSymbol)
         {
-            JToken token = await MakeJsonRequestAsync<JToken>("/products/" + symbol + "/ticker");
-            return ParseTicker(token, symbol);
+            JToken token = await MakeJsonRequestAsync<JToken>("/products/" + marketSymbol + "/ticker");
+            return ParseTicker(token, marketSymbol);
         }
 
         protected override async Task<IEnumerable<KeyValuePair<string, ExchangeTicker>>> OnGetTickersAsync()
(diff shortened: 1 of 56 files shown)


superhappychris:
@bichuga you have it backwards ... another way of thinking about it when you're placing an order is that the price is always denominated in the quote currency, and the quantity is always the quantity of the base currency. I know it can be confusing, especially because a couple of the exchanges even get it wrong (Bittrex :man_facepalming:), but these are the standard terms used when trading [currency pairs](https://en.wikipedia.org/wiki/Currency_pair). Trying to come up with new names will confuse people even more I'm afraid

jjxtra:
We can update the readme to be clear about naming conventions with whatever we decide is right.

bichuga:
Are there any other names for this that would be so clear as to not
necessitate docs? Altcoin or altcurrency on one side seems pretty strong to
me. Anything else?

superhappychris:
@bichuga I see a couple problems with using BaseCurrency/AltCurrency...

- For the vast majority of exchanges (I think Bittrex is the only exception), the term BaseCurrency means the opposite of how you want to use it. Obviously the path of least confusion would be to go with what the majority of exchanges do.  
- AltCurrency or AltCoin already has a pretty specific meaning to most people and it doesn't have anything to do with trading currency pairs, it has to do with the subjective quality of the coin or token in question.  What you consider a shitcoin may not be considered a shitcoin by everyone else, and those definitions may change over time.
- Honestly, the only reason we'd need to be clear about the terms BaseCurrency / QuoteCurrency is to clarify to contributors that a couple rogue exchanges use them incorrectly.  Other than that, those terms would only be confusing to a person that is new to the idea of trading currency pairs in general, or that got some bad info from somewhere.  Those people are going to be confused no matter what terms we use.  Nothing wrong with that, I was confused at first too, everyone has to learn at some point.

I don't think we need to reinvent the wheel here.  Base Currency / Quote Currency are the most widely used terms out there. The majority of exchanges use those terms.

bichuga:
I get your point about not wanting to imply a quality or connotation with
the name. Yet your point that other folks use these terms doesn't mean they
are good choices. The fact that multiple major exchanges with all the API
reviews involved could have gotten these names exactly backwards
demonstrates the quality of the naming scheme.

On exchanges the tickers are organized as a few central currencies
branching to many. What about HubCurrency/SpokeCurrency?

superhappychris:
@bichuga well there's only one major exchange I know of that actually got the meaning of the terms backwards, and that's Bittrex. Given the fact that Bittrex was created by a couple of developers with a background in security and not finance, it's understandable how they could have gotten it wrong, and fixing a mistake like that would be very costly.  I'm trying to help us avoid making the same mistake 😄 

I guess we'll just have to agree to disagree that BaseCurrency / QuoteCurrency are good names.  Not only do they provide users the information they need to place an order for that trading pair (price in quote currency, quantity in base currency), but they are also the industry standard terms which means using them will lead to less confusion overall.

vslee:
I like BaseCurrency / QuoteCurrency because this will also align with other APIs as well as when viewing the currencies on the websites of the exchanges. Inventing new terms would just make it more difficult for the users to have to learn new concepts.

bichuga:
If everyone likes it but me, just make the documentation clear 😃

Also super double check the ones that got flipped are actually correct now

jjxtra:
Casting my vote for base / quote currency

superhappychris:
@jjxtra if you take a look at how the exchanges use "symbol" in their apis, they all use it in the currency pair sense, not the single asset sense, which is why I went with the former when refactoring.  To me, this is another case of why come up with a new term when one already exists that is used by the exchanges?

jjxtra:
Right. Maybe we should rename the new "currency" parameters to "currencyPair" parameters where they used to be "symbol" parameters.

jjxtra:
Nevermind, it looks like the methods that take currency/market pairs are still using the "symbol" parameter name, such as GetTickerAsync.

vslee:
Market or CurrencyPair are good. Symbol is vague but is used in some exchanges, but doesn't seem like it would hurt to avoid it. Perhaps Symbol can be placed in the doc comments so that it does appear if ppl mouse over Market/CurrencyPair.

superhappychris:
I think market is too vague, especially since we already have a class called ExchangeMarket.  If we go with market, then the ExchangeMarket class will have a property called Market that is of type string, but there will be other places that refer to a "market" as an ExchangeMarket type...

What if we go with "MarketSymbol"?  It's unambiguous, more in line with what the many of the exchanges use terminology wise, and lends itself to future changes if/when we decide to create a MarketSymbol type to encapsulate the metadata that can't be captured in a string (individual currency names, separator character, order, etc.).  Thoughts?

jjxtra:
Sounds reasonable. How do you feel about renaming all the symbol params in the pull request? Hopefully just a case sensitive search and replace of entire word.

superhappychris:
@jjxtra I can take care of it 👍

jjxtra:
If we do use a MarketSymbol class in the future it's easy enough to make an implicit string operator conversion, so people can still pass string parameter for this.

jjxtra:
I did some code edits on my iPad, but obviously wasn't able to ensure they compile, let me know if you see any issues.

superhappychris:
Ok it builds again

jjxtra:
Anyone have any objection to merging this?

vslee:
Go for it

PR #280, opened 2018-10-27
Add more functionalities to Bistamp and Coinbase API

atp4al:
- Add the possibility to send market orders with Bitstamp
- Add limit = 1000 to retrieve more BitstampTransaction
- Add a way to retrieve Coinbase Fills


jjxtra:
Please don't add synchronous end points, you can simply append .Sync() to an Async end point to do that. 

I performed some other cleanup. 

Thanks for sharing.

atp4al:
Thanks for your corrections and for merging.

PR #286, opened 2018-11-01
Fix bad argument when creating KeyValuePair objects inside OnGetTickersWebSocket

superhappychris:
Somehow I missed this when I did the naming refactor


vslee:
Fixes #267

PR #287, opened 2018-11-01
Fix parsing of marketSymbol and Id in Kucoin trade stream.

vslee:
In Kucoin, the trade Id and order Id are long hexadecimals, such as 5bda8e5c9dda157958b7b4cc. Thus, since ExchangeTrade.Id is a long, the Id cannot be parsed correctly. If one day ExchangeTrade.Id is converted to string instead, then the line can be uncommented and the oid can be parsed.


vslee:
related to #285

PR #310, opened 2018-11-27
Update ExchangeOkexAPI.cs

igolenastov:
"ltc_btc"  BaseCurrency = pieces[1]


vslee:
I agree that these are backwards

PR #313, opened 2018-12-04
Added exchange id to order book, order request and order result

Lipindim:
Allow create aggregated order book, send order to different exchange not using additional params


jjxtra:
It's not clear to me what ExchangeId is. Is it a name? Unique id?

Lipindim:
> It's not clear to me what ExchangeId is. Is it a name? Unique id?

This is the name of the exchange and it is unique.

PR #325, opened 2018-12-17
Updated ParseOrder function to read marketSymbol from product_id

yatendra:
Updated ParseOrder function to read marketSymbol from product_id instead of id. Updated TargetFramework in ExchangeSharp.cspoj to move nestandard2.0 before net472 to be able to support intellisense in netcore projects also.


yatendra:
Thanks. That works.

PR #326, opened 2018-12-17
Updating binance get tickers

Peixer:
# Changes
- Updating Binance GetTicker, to nomarlize symbol before request.
- Updating Bitfinex exchange GetAmounts, adding new parameter to choose wallet.


jjxtra:
Looks like there are conflicts. You should pull latest and re-sync. Normalize symbol is gone and not necessary in 99% of cases.

Peixer:
this change was necessary in my application, that is in production mode.

jjxtra:
OnGetTickerAsync will have already had NormalizeMarketSymbol called. Do not call OnGetTickerAsync directly, call GetTickerAsync.

Peixer:
Left this change
- Updating Bitfinex exchange GetAmounts, adding new parameter to choose wallet.

PR #331, opened 2018-12-26
 Added ability to send non-post-only limit orders to Coinbase pro using preset order.ExtraParameter["post_only"]

yatendra:


ExchangeSharp/API/Exchanges/Coinbase/ExchangeCoinbaseAPI.cs
```diff
@@ -574,7 +574,9 @@ protected override async Task<ExchangeOrderResult> OnPlaceOrderAsync(ExchangeOrd
             switch (order.OrderType)
             {
                 case OrderType.Limit:
-                    payload["post_only"] = "true";
+                    // set payload["post_only"] to true for default scenario when order.ExtraParameters["post_only"] is not specified
+                    // to place non-post-only limit order one can set and pass order.ExtraParameters["post_only"]="false"
+                    payload["post_only"] = (order.ExtraParameters["post_only"] == null) ? "true" : order.ExtraParameters["post_only"];
```
jjxtra: I would suggest trygetvalue here

ExchangeSharp/Model/ExchangeOrderRequest.cs
```diff
@@ -70,7 +70,7 @@ public class ExchangeOrderRequest
         /// Not all exchanges will use this dictionary.
         /// These are added after all other parameters and will replace existing properties, such as order type.
         /// </summary>
-        public Dictionary<string, object> ExtraParameters { get; private set; } = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+        public Dictionary<string, object> ExtraParameters { get; set; } = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
```
jjxtra: Why does this need a public set?

yatendra: This needs to be set in the ExchangeOrderRequest object being passed to SubmitOrder function from the application code outside this framework, ExtraParameters attribute needs to have public set

jjxtra: My preference would be to have everyone use .CopyTo on their dictionary to copy the items to this dictionary, to ensure that any params in there are not blown away. If you want to blow away the parameters already in there, you can clear the dictionary explicitly before calling CopyTo on your dictionary. Anyone else thoughts?

``` myDictionary.CopyTo(orderRequest.ExtraParameters); ```

jjxtra: This also avoids null ref exceptions until we implement the new C# 8.X null handling.

PR #332, opened 2018-12-27
Added ability to send non-post-only limit orders to Coinbase pro using preset order.ExtraParameter["post_only"]

yatendra:
Added ability to send non-post-only limit orders to Coinbase pro using preset order.ExtraParameter["post_only"]


yatendra:
I closed previous pull request and created new one with change you recommended. About your question about why we need ExchangeOrderRequest.ExtraParameters to have public set, it needs to be set in the ExchangeOrderRequest object being passed to SubmitOrder function from the application code outside this framework, so ExtraParameters attribute needs to have public set

yatendra:
CopyTo method works for me. I will revert this change. Thanks.

Issue #335, opened 2019-01-02
Cannot send a content-body with this verb-type for BitMEX GET


christso:
I have a BUG REPORT.

**Details:**

When I call the method GetOpenOrderDetailsAsync() or GetAmountsAsync() within ExchangeBitMEXAPI, I get the error:

```
System.Net.ProtocolViolationException
  HResult=0x80131509
  Message=Cannot send a content-body with this verb-type.
  Source=System.Net.Requests
  StackTrace:
   at System.Net.HttpWebRequest.InternalGetRequestStream()
   at System.Net.HttpWebRequest.BeginGetRequestStream(AsyncCallback callback, Object state)
   at System.Net.WebRequest.<>c.<GetRequestStreamAsync>b__67_1(AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl(Func`3 beginMethod, Func`2 endFunction, Action`1 endAction, Object state, TaskCreationOptions creationOptions)
   at System.Threading.Tasks.TaskFactory`1.FromAsync(Func`3 beginMethod, Func`2 endMethod, Object state)
   at System.Net.WebRequest.<GetRequestStreamAsync>b__67_0()
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at ExchangeSharp.APIRequestMaker.InternalHttpWebRequest.<WriteAllAsync>d__14.MoveNext() in C:\Users\chris\source\repos\ExchangeSharp\ExchangeSharp\API\Common\APIRequestMaker.cs:line 95
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #336, opened 2019-01-03
Skip writing of empty content-body for GET requests

christso:
Skip writing of empty content-body for GET requests. This fixes the error in BitMEX GET requests - "Cannot send a content-body with this verb-type".

Fixes issue #335.


jjxtra:
Thanks for sharing!

PR #349, opened 2019-02-16
Support Bitbank.cc

joemphilips:
BitBank.cc is one of the most popular exchange in japan.
https://bitbank.cc/
I want this in my own project for trading JPY.
Private methods are manually tested.


jjxtra:
Wow this is a lot of work, thanks for sharing

rockstardev:
Great job Joe!

PR #357, opened 2019-03-04
Support GetTickers in ExchangeBitBankAPI and some bugfix

joemphilips:
BitBank has recently supported a new api for fetching all the prices at once.


joemphilips:
May I ask when the new version will be released? Is there a document about release process somewhere?

Issue #350, opened 2019-02-20
Kucoin API has been updated


kcabrera99:
The Kucoin API was updated recently and now the API calls aren't working.

jjxtra:
You are welcome to submit a pull request if you would like to fix this

username77:
> The Kucoin API was updated recently and now the API calls aren't working.

try to use this api: https://github.com/mscheetz/KuCoinApi.Net/tree/v2.0

jjxtra:
If someone wants to create a pr with that KuCoinApi.Net project I will happily merge it in

kcabrera99:
I'm trying to do a pull request on exchangesharp but it's not letting me.

username77:
> I'm trying to do a pull request on exchangesharp but it's not letting me.

post the code here

vslee:
To create a pull request, you have to fork the repository first. Then on your fork, create a new branch and commit your changes there. Then, go back to github and view your fork, there will be a button to create a pull request. Here are detailed instructions (but it doesn't have to be as complicated as these instructions) https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github

Or, if you want, you can just post the new code here on this thread and someone will probably create the pull request for you.

vslee:
I created the first pull request to help you get started

kcabrera99:
I was able to update the getcandles method but I'm having trouble with the private apis.  Getting an error about the signature.  "Invalid KC-API-SIGN"

kcabrera99:
I think I got it now

kcabrera99:
Got most of the gets working, but can't get past the "Invalid KC-API-SIGN" errors for the POST requests.

vslee:
It likely has something to do with signing the message: https://docs.kucoin.com/#signing-a-message

kcabrera99:
I've been looking at that all day trying to get it to work.

vslee:
Okay, just make a pull request with what you have so far and I'll take a look

kcabrera99:
OK, I think I created the pull request.  I was able to get the post requests working for place order.  I had to change the contenttype to application/json.

I have tested getcandles, placeorder, getorderdetails which calls getopenorders and getcompletedorders.  

Kucoin now offers a way to get the details for a specific orderID but I did not have time to implement that.  Also, I changed the getnonceoffset to use a timestamp function that they now offer instead of getting the tickers. 

[...]

PR #361, opened 2019-03-09
Kucoin API2 minor updates

username77:
OnGetAmountsInternalAsync has changed on api v2
OnGetTickerAsync has changed on api v2
ParseTickers added
OnGetAmountsInternalAsync has changed on api v2 (now are multi accounts) to read amount for only trading account


vslee:
Progress toward #350

PR #366, opened 2019-03-23
BitFinex trade websocket snapshot

szmcdull:
GetTradesWebSocket() calls callback when snapshot arrives on subscription success.

I further suggest adding some flags to ExchangeTrade class to indicate whether a trade is from snapshot, and whether a trade is the last one in the snapshot.


jjxtra:
It is pushed to master, ExchangeTradeFlags enum, Flags property on ExchangeTrade. IsBuy merged into flags, along with snapshot and last snapshot flags available.

PR #383, opened 2019-05-07
Okex - ExchangeMarket.MaxTradeSize fixed

alperencgmn:
MaxTradeSize is not marketSymbolToken and ConvertInvariant sets maxTradeSize always 100 so it disabled


jjxtra:
Shouldn't it be cast to a decimal before doing the math pow function?

jjxtra:
Looks like Math.Pow is double only so I guess it does not really matter.

PR #386, opened 2019-05-12
Binance Implement GetHistoricalTradesAsync based on Id's

ParadoxForge:
Binance's API allows for fetching historical data based on ID's.

This method is much handier, reliable and efficient in syncing data than calls done by date.

I will be attempting to implement this is more exchanges at a later stage.

### DIFF
--- a/ExchangeSharp/API/Exchanges/Binance/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/Binance/ExchangeBinanceAPI.cs
@@ -359,6 +359,50 @@ namespace ExchangeSharp
             await state.ProcessHistoricalTrades();
         }
 
+        protected override async Task OnGetHistoricalTradesAsync(Func<IEnumerable<ExchangeTrade>, bool> callback, string marketSymbol, long startId, long? endId = null)
+        {
+            /* [ {
+            "a": 26129,         // Aggregate tradeId
+		    "p": "0.01633102",  // Price
+		    "q": "4.70443515",  // Quantity
+		    "f": 27781,         // First tradeId
+		    "l": 27781,         // Last tradeId
+		    "T": 1498793709153, // Timestamp
+		    "m": true,          // Was the buyer the maker?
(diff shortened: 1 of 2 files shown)


jjxtra:
Just getting thoughts from everyone about whether we can keep ExchangeAPI to one set of historical methods. Maybe we change signature to a class for the filter parameters? The class could have start/end date and start/end id, and anything else we need to add in the future?

vslee:
How would you know which filter paramters work for each exchange?

jjxtra:
Each exchange API could throw not supported or invalid operation exception, beyond that probably it would be assumed begin/end dates would work, and then ids would depend on exchange. Future parameters, the same.

vslee:
Ya, I suppose there's no good way to specify / check at compile time

ParadoxForge:
This is why i opted for implementing both functions, in the vast majority of use-cases users of this lib will only implement 1 exchange, in which case choosing the correct function is easy, in other cases the applications complexity will be enough to warrant something in the line of a switch to pick the correct function at run-time.

But that's just my 2 cents.

jjxtra:
Do you know of any other exchange that uses ids? We could consider removing it from the public interface and just exposing it on the binance exchange class...

ParadoxForge:
You have a valid point, No I do not.

I will add that I have not gotten round to checking their API's to that extent yet.

@vslee What do you think?

vslee:
I think adding it just on the Binance class would be a good start. If other exchanges get support, then it can be made into a general method. That way, there is less frustration for users who code up an app, only to see that it doesn't work when they finally run it.

jjxtra:
Agreed. Please add a TODO to the method to revisit in the future for other possible exchanges.

jjxtra:
Happy to pull this in, please just revert the interface change for now.

ParadoxForge:
Sure, I'll do so this weekend

ParadoxForge:
@jjxtra PR Updated, had some time this evening

jjxtra:
Thanks for taking the time!

PR #388, opened 2019-05-16
Refined Bittrex API RateLimit

ParadoxForge:
Bittrex does not have a configured RateLimit, which uses the base 5calls/15seconds.

We can see from their documentation that the accepted rate limit is 60 calls / 60 seconds


jacobdeboer33:
Hi there

I used the bittrex api with the added rategate line, but it still bans me when I try to get candles for every BTC market in a loop. It bans me after about 20 markets. So I think the rate limit not correct

Thanks,
Jacob

jjxtra:
Feel free to change the rate gate to your own limit.

PR #392, opened 2019-06-05
Added InvariantCulture for use in enviroments with other cultures

edalmasso:
Problem: when requesting the order book on Livecoin and the exchange returns for example: asks = {["7800.07", "0.00172",155975495],["7.93E+3","0.00018984","155975362"]} 
If using a culture different than "us", the conversion of "7.93E+3" throws an error as it is not recognized (it only recognizes "7,93E+3").
Setting CultureInfo.InvariantCulture forces to use always the same culture on convertions.

### DIFF
--- a/ExchangeSharp/API/Exchanges/_Base/ExchangeAPIExtensions.cs
+++ b/ExchangeSharp/API/Exchanges/_Base/ExchangeAPIExtensions.cs
@@ -313,6 +313,7 @@ namespace ExchangeSharp
             int maxCount = 100
         )
         {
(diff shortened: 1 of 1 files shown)


edalmasso:
If you think the line should be placed in another place to affect all convertions, please move it to where it should go.
This is the line that fails: 
`array[0].ConvertInvariant<decimal>()`

jjxtra:
Fixed in da5386473ddb32443b54cb209a83003b4e5b1dfc

No need to hack around it, fix problem at source.

PR #393, opened 2019-06-05
Fix on ProcessRequestAsync as it was not able to open a new order

edalmasso:
On open order attempt the server sent error: "invalid sign"
Fixed payloadForm format. 
The "**/**", "**@**" and "**;**" should be replaced only for Livecoin.
From official documentation: [https://www.livecoin.net/api/examples#c_sharp](https://www.livecoin.net/api/examples#c_sharp)
Copied the same method from API Documentation.

### DIFF
--- a/ExchangeSharp/API/Exchanges/Livecoin/ExchangeLivecoinAPI.cs
+++ b/ExchangeSharp/API/Exchanges/Livecoin/ExchangeLivecoinAPI.cs
@@ -36,6 +36,7 @@ namespace ExchangeSharp
             if (CanMakeAuthenticatedRequest(payload))
             {
                 string payloadForm = CryptoUtility.GetFormForPayload(payload, false);
+                payloadForm = http_build_query(payloadForm);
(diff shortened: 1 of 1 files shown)


jjxtra:
I checked in something, let me know if it works.

edalmasso:
I am getting errors on other Livecoin methds like: OnGetAmountsAvailableToTradeAsync
Error is: _One or more errors occurred. (Cannot send a content-body with this verb-type.)_
As you said, its Livecoin weirdness...
Also tested on Bitfinex and Bittrex and both works ok.

jjxtra:
Sounds like livecoin code needs some work I guess... were those methods working before I changed the form method?

edalmasso:
Yes, if you use the fork I sent, all methods works ok.
I will be using Livecoin a lot, so I will send new PR if needed.
I am still learning how everything works, so my fixes will be very specific (you can then move them to the correct places if needed)

jjxtra:
Ok just update this pr to undo my changes to the livecoin file

PR #401, opened 2019-07-17
Convert TradeID in trade stream to string from long

vslee:
Exchanges use a varying set of types for Trade IDs. Currently, we use a bunch of ugly hacks to convert them to long in the trade stream. In BitMex, we were attempting to parse a GUID into a long and thus throwing and then catching an exception for each trade. In Huobi, we were setting all of the trade IDs to that of the parent tick. Finally, in many other exchanges, were substituting the timestamp ticks for the trade ID as we didn’t know what else to do.

These hacks have a variety of problems. First, some of them do not uniquely identify trades, such that multiple trades can have the same ID. Also, since we are not using the original trade IDs from the exchanges, the new IDs are not able to be used to look up trades from those same exchanges.

Changing the type to string allows the end user to access the actual Trade ID. They can then parse them to long, hash them, or ignore them and use timestamp. This would be a breaking change. However, it would be a small one and allows them to make decisions about the tradeoffs themselves.


jjxtra:
Thanks for taking the time.

PR #404, opened 2019-07-20
Bitstamp: added Trade stream (websocket)

vslee:


jjxtra:
Did you make any code changes to crypto utility file? It looks like just whitespace from what I can tell.

vslee:
Added UnixTimeStampToDateTimeMicroseconds()

PR #406, opened 2019-07-22
allow async callbacks in GetTradesWebSocket

vslee:
- may also consider doing the same for all of the other GetxxxWebSocket methods


jjxtra:
You're on a roll! Thanks for sharing these changes, very generous of you.

PR #408, opened 2019-07-24
Changes Coinbase Trade stream to 'matches' from 'ticker'

vslee:
- doing so provides additional information such as MakerOrderId and TakerOrderId
- created new subclasses of ExchangeTrade to provide this info
- does the same for BinanceAggregateTrade as well
- does all of this in a backwards compatible way (does not break existing code)
- now logs Trade steam errors in Coinbase

### DIFF
--- a/ExchangeSharp/API/Exchanges/Binance/ExchangeBinanceAPI.cs
+++ b/ExchangeSharp/API/Exchanges/Binance/ExchangeBinanceAPI.cs
@@ -294,7 +294,7 @@ namespace ExchangeSharp
 
                 // buy=0 -> m = true (The buyer is maker, while the seller is taker).
                 // buy=1 -> m = false(The seller is maker, while the buyer is taker).
-                await callback(new KeyValuePair<string, ExchangeTrade>(marketSymbol, token.ParseTrade("q", "p", "m", "E", TimestampType.UnixMilliseconds, "a", "false")));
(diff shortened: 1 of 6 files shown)


vslee:
This does end up unsealing ExchangeTrade, but I can't see an easy way of implementing this without doing so.

jjxtra:
It looks ok. The only thing I would request is to put the coinbase and binance trade classes into their own namespace (ExchangeSharp.Coinbase and ExchangeSharp.Binance) and put the them in files in the sub-folder for those exchanges.

Thank you for sharing!

jjxtra:
If you wanted to do it without unsealing, you could make the trade classes not inherit exchange trade but instead contain an exchange trade object. But I don't have a problem unsealing it.

vslee:
The problem with having them inherit from a different ExchangeTrade class is that it then breaks the unified experience where ppl can use an ExchangeTrade in the callback without caring about which exchange it came from.

vslee:
I moved the new classes into their own namespace. I realized I misread your second comment. I'm not sure how things would work if these new classes contained an ExchangeTrade instead of inheriting from it. The signature of GetExchangeTrades is:
 ```public virtual IWebSocket GetTradesWebSocket(Func<KeyValuePair<string, ExchangeTrade>, Task> callback, params string[] marketSymbols)```c#
Thus, it returns an ExchangeTrade for all of the exchanges. But if Binance and Coinbase suddenly started returning a BinanceAggregateTrade that did not inherit from ExchangeTrade, then BinanceAggregateTrade could not be passed through that callback. Is that what you had in mind?

jjxtra:
Thanks for sharing this, sorry I didn't get it merged in sooner...

PR #417, opened 2019-08-13
OKex: fix market symbol bugs

### LINKED ISSUE #402: Okex GetOrderBookWebSocket only works for swap symbols
On [this line](https://github.com/jjxtra/ExchangeSharp/blob/360e0e069e2dca9dcd89e19e8ca8c64457b02f14/ExchangeSharp/API/Exchanges/Okex/ExchangeOkexAPI.cs#L239) you can see that Okex orderbook WS-channel is created as "swap/depth:{0}-SWAP", which only works for swap symbols. For spot symbols you need to use "spot/depth:{0}". 
In [the recent PR](https://github.com/jjxtra/ExchangeSharp/pull/400) "swap" is simply replaced with "spot" in another method, should I do the same? Or it's better to make a common solution, for example we can use {0}-SWAP as symbol string for swap symbols and the code will relplace "spot/" to "swap/" depending on the existense of -SWAP ending

### LINKED ISSUE #411: OKex - wrong symbol separator
GetTickersAsync() gives following exception...

System.InvalidOperationException: Splitting Okex symbol 'wxt_btc' with symbol separator '-' must result in exactly 2 pieces.

Looks like the symbol separator should be '_' (instead of '-') 

image

vslee:
- upgraded OnGetMarketSymbolsMetadataAsync() to use V3 api
- log websocket connect error instead of throwing an exception
- changed Trade websocket parsing to use ExchangeAPIExtensions
- upgraded OnGetTickersAsync() to use V3 api
- upgraded OnGetTickerAsync() to use V3 api
- overrode ExchangeMarketSymbolToCurrencies() to to handle futures and swap symbols
- fixes #402, fixes #411

### DIFF
--- a/ExchangeSharp/API/Exchanges/Okex/ExchangeOkexAPI.cs
+++ b/ExchangeSharp/API/Exchanges/Okex/ExchangeOkexAPI.cs
@@ -26,10 +26,10 @@ namespace ExchangeSharp
         public string BaseUrlV3 { get; set; } = "https://www.okex.com/api";
         public override string BaseUrlWebSocket { get; set; } = "wss://real.okex.com:10442/ws/v3";
 
-	/// <summary>
-	/// China time to utc, no DST correction needed
-	/// </summary>
-	private static readonly TimeSpan chinaTimeOffset = TimeSpan.FromHours(-8);
+		/// <summary>
+		/// China time to utc, no DST correction needed
+		/// </summary>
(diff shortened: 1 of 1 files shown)


vslee:
Keeping this PR open for a day in case there are any comments/reviews from @jjxtra, @vlebedev7, or @GonzoKK

PR #425, opened 2019-08-20
Added user data stream (Binance)

RapidEU:
Added user data functionality
https://github.com/binance-exchange/binance-official-api-docs/blob/master/user-data-stream.md

UserData:
The stream is used to get notifications on changed, whether it's orders you make or account changes that are happening, this is a good way to monitor your trade status on new event.

You must first get your ListenKey and then send it in order to open a new UserData websocket.

I'm sorry for the mass indentation, it's automatic.
Check it carefully however, I'm here for rejects.

Thanks for the hardwork you put into this project, jjxtra and vslee.

The first commit name has nothing to do with Kraken however, it's from an old PR I made.


ExchangeSharp/API/Exchanges/Binance/Models/UserDataStream.cs
```diff
@@ -0,0 +1,164 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace ExchangeSharp.API.Exchanges.Binance.Models
```
vslee: Can change this namespace to just ExchangeSharp.Binance

jjxtra:
Is the new exchange interface method for the key a Binance only thing?

RapidEU:
I think it is actually. So it could be better just to make it a public method inside the Binance API, or another way is to just create a method that generates the listenkey when the user tries to create userdata stream and just define a property that holds it for further interaction
Thoughts?

jjxtra:
Yea let's make it a public Binance method for now. Can revisit if another exchange adds similar functions.

vslee:
Needs rebase but otherwise looks okay to me.

PR #428, opened 2019-08-29
Fix Kraken problems with coin names

edalmasso:
Fixed problems with symbol names and names conversions.
Enable Kraken GlobalSymbolTest.

The code works, but I an not sure if it is in the correct place or can be optimized.


vslee:
Looks reasonable to me, but I'll let @jjxtra comment on this one as I haven't worked in this section of the code much.

jjxtra:
Good enough for now :)

PR #433, opened 2019-09-05
New exchange Digifinex

szmcdull:


ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
```diff
@@ -0,0 +1,514 @@
...
+                //    "BTC_USDT"
+                //  ],
+                //  "id": null
+                //}
+                JToken token = JToken.Parse(CryptoUtility.DecompressDeflate((new ArraySegment<byte>(msg, 2, msg.Length-2)).ToArray()).ToStringFromUTF8());
+                if (token["method"].ToStringInvariant() == "depth.update")
+                {
+                    var args = token["params"];
+                    var data = args[1];
+                    var book = new ExchangeOrderBook { LastUpdatedUtc = DateTime.UtcNow, MarketSymbol = (string)args[2] };
```
vslee: Try using CryptoUtility.UtcNow instead of DateTime.UtcNow for mocking later

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
```diff
@@ -0,0 +1,514 @@
...
+                        if (clean)
+                            flags |= ExchangeTradeFlags.IsFromSnapshot;
+                        if (i == x.Count - 1)
+                            flags |= ExchangeTradeFlags.IsLastFromSnapshot;
+
+                        await callback.Invoke(new KeyValuePair<string, ExchangeTrade>(
+                            symbol, new ExchangeTrade
+                            {
+                                Id = (string)trade["id"],
+                                Timestamp = DateTimeOffset.FromUnixTimeSeconds(0).AddSeconds((double)trade["time"]).LocalDateTime,
```
vslee: There is a method CryptoUtility.UnixTimeStampToDateTimeSeconds(double) that can help you do this conversion but keep the timestamp in UTC.

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
```diff
@@ -0,0 +1,514 @@
...
+        {
+            Dictionary<string, object> payload = await GetNoncePayloadAsync();
+            JToken token = await MakeJsonRequestAsync<JToken>($"/spot/order?order_id={orderId}", payload: payload);
+            var x = token["data"];
+            return new ExchangeOrderResult
+            {
+                MarketSymbol = (string)x["symbol"],
+                OrderId = (string)x["order_id"],
+                OrderDate = DateTimeOffset.FromUnixTimeSeconds((long)x["created_date"]).LocalDateTime,
+                FillDate = DateTimeOffset.FromUnixTimeSeconds((long)x["finished_date"]).LocalDateTime,
```
vslee: There is a method CryptoUtility.UnixTimeStampToDateTimeSeconds() that can help you do this conversion but keep the timestamp in UTC.

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
```diff
@@ -0,0 +1,514 @@
...
+                    timeWindow = (timeFaster.Negate().TotalSeconds*10).ToString();
+                    NonceOffset = TimeSpan.FromSeconds(2.5);
+                }
+                else
+                    NonceOffset = now - serverDate; // how much time to substract from Nonce when making a request
+            }
+            catch (Exception)
+            { }
+        }
+
```
jjxtra: I've added a checkin just now that implements a OnGetNonceOffset method and adds 3 new properties for the exchange to implement if the api supports a server timestamp:

NonceEndPoint, NonceEndPointField, NonceEndPointStyle

Can you explain why you have logic to divide by 2, multiply by 10 and then use a 2.5 second offset? What is timeWindow? I'd like to see if we can move some of this into the base OnGetNonceOffset method to avoid future exchanges needing this kind of logic.

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
szmcdull: Oh I thought local time is more helpful in some situations

jjxtra: If any end user wants local time, they can call the appropriate method on the DateTime object which will be in UTC by default and easily able to do the conversion.

vslee: Having different timezones in different places introduces inconsistencies and may lead to inadvertent bugs in the user's code

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
szmcdull: / 2 is to take the average of send and recv transfer overhead and estimate the time when the server take its time.

The timeWindow is a parameter to tell the Digifinex server how much time span it can tolerate that the client time is behind of its server time. However, the server cannot tolerate that client time is ahead of server time. When NonceOffset is negative, I reset it to +2.5 seconds to prevent ExchangeSharp to send nonce that ahead of the server time. And I use the timeWindow to let the server tolerate the client time behind. x10 is just a number that is kind of *safe* (see the [doc](https://docs.digifinex.vip/en-ww/v3/) and search for `ACCESS-RECV-WINDOW`).

When the client time is ahead of the server time (NonceOffset > 0), I just follow the usual process flow of ExchangeSharp.

jjxtra: Thanks for the clarification. We'll have to see in the future if we can move this into the base class, but for now this method looks good.

ExchangeSharp/API/Exchanges/Digifnex/ExchangeDigifinexAPI.cs
jjxtra: Thanks so much for changing to utc, can you switch all the (long) and (string) casts to use ConvertInvariant<long>() and ToStringInvariant()? This is the standard way to convert longs and strings to avoid any chance of culture parsing and formating issues. We are close now!

szmcdull:
I found that most other exchange use uppercase symbols. Is it too late to commit a mod that make symbols to uppercase?

PR #448, opened 2019-09-20
new nuget 0.6.1

vslee:


jjxtra:
I guess we don't need our summary hack anymore since it is deprecated :)

PR #455, opened 2019-10-01
Fixed ClientWebSocket disposal

vslee:
- cancel CTS and set 'disposed' bool after attempting to close connection
- dispose of the webSocket
- mark messageQueue as complete
- close MemoryStream in ReadTask()

### DIFF
--- a/ExchangeSharp/Utility/ClientWebSocket.cs
+++ b/ExchangeSharp/Utility/ClientWebSocket.cs
@@ -237,10 +237,10 @@ namespace ExchangeSharp
         {
             CreateWebSocket();
 
-            // kick off message parser and message listener
-            Task.Run(MessageTask);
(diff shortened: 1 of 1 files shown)


jjxtra:
Only question I have is can we ensure Dispose is only called once? What did you run into having dispose set to true and checked before doing the task.run?

vslee:
If the either the cancellationToken is cancelled, or the disposed is set to true first, then the connection may be prematurely disconnected prior to the task.run running. And thus, the CloseAsync/CloseOutputAsync is not always run.

vslee:
In terms of Dispose() being called more than once, it would just spawn another Task, but then everything inside would be okay. It would check to see if the socket is open, and if not, then it wouldn't try to close it again. And the last three lines can be called multiple times w/o issue.

PR #470, opened 2019-10-22
BL3P Websockets (Orderbook) implementation and tests

johnnyasantoss:
Add Orderbook via WebSockets

Made OnGetMarketSymbolsMetadataAsync protected internal to be able to mock in tests

Add unit tests and timestamp

Rename folder and namespace

### DIFF
--- a/ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
+++ b/ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
@@ -2,6 +2,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using ExchangeSharp.API.Exchanges.BL3P;
+using ExchangeSharp.API.Exchanges.BL3P.Models;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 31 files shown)


johnnyasantoss:
This implementation can be tested with the command `dotnet run -- ws-orderbook -e bl3p -m btceur` in the `ExchangeSharpConsole` folder.

vslee:
You can update the readme to show the new websockets feature

johnnyasantoss:
Rebased and updated the readme :smiley:

vslee:
Somewhat off topic, but do you have sway with BL3P? I was looking at their websockets API and noticed that the [trade stream](https://github.com/BitonicNL/bl3p-api/blob/master/docs/public_api/websocket.md#21---trades) doesn't include a trade ID (or order ID). Is that something you could have them add? Would be helpful in figuring out if there are missing/duplicate trades. 
Also most exchanges give a side (Buy/Sell) for each trade. So that would be helpful to add as well (but less important).

jjxtra:
Id still like the bl3p exchange specific model objects to be marked as internal instead of public, anyone have any strong objections?

vslee:
I can change it to private nested classes like the NDAX

johnnyasantoss:
> I'd still like the bl3p exchange specific model objects to be marked as internal instead of public, anyone have any strong objections?
> 
> In my ideal world they would simply be child objects inside the partial exchange class, that way we keep the namespace clean, but I am less picky about that.

I would prefer to keep them in separate files and follow the Microsoft Csharp standards. But I agree that there's no need for these classes to be public :) I'll fix it in my next PR.

johnnyasantoss:
> Somewhat off topic, but do you have sway with BL3P? I was looking at their websockets API and noticed that the [trade stream](https://github.com/BitonicNL/bl3p-api/blob/master/docs/public_api/websocket.md#21---trades) doesn't include a trade ID (or order ID). Is that something you could have them add? Would be helpful in figuring out if there are missing/duplicate trades.
> Also most exchanges give a side (Buy/Sell) for each trade. So that would be helpful to add as well (but less important).

I work at [Bitonic](https://bitonic.nl/en/) (the company behind BL3P) and I can check with our team about those additional IDs :)

johnnyasantoss:
> > I'd still like the bl3p exchange specific model objects to be marked as internal instead of public, anyone have any strong objections?
> > In my ideal world they would simply be child objects inside the partial exchange class, that way we keep the namespace clean, but I am less picky about that.
> 
> I would prefer to keep them in separate files and follow the Microsoft Csharp standards. But I agree that there's no need for these classes to be public :) I'll fix it in my next PR.

Regarding the PR #474 I would like to point out that using this strategy of using partial classes we are creating yet another indentation level and at the same time making the classes a new type of name space, which is bad for the type itself, because it gets polluted with symbols that could be hidden by access modifier.
Should we keep with the same strategy or start using `private internal` for model types and other helper types?

Issue #450, opened 2019-09-24
C# 8.0 / .NET CORE 3


jjxtra:
The csharp8 branch contains all c# 8 work including nullable reference types, pattern matching, async enumerable and other syntax shortcuts. Resolve this issue upon merge to master with no warnings.

vslee:
I took a look at the first commit. When I try running it, upon the instantiation of the exchangeAPI, it throws: `System.NotSupportedException: 'The requested security protocol is not supported.'`

vslee:
nevermind, looks like it may be something on my end

jjxtra:
Nice, get rid of ssl 3 looks like a good idea to me.

vslee:
We might want to merge the changes so far into the master branch before this branch gets stale. Otherwise, would probably have conflicts in the future.

jjxtra:
I may end up deleting the branch. There’s a lot to do to really get on csharp 8 and I just don’t have the time to do it.

vslee:
I haven't looked much into C# 8, but my understanding is that its features can be introduced gradually rather than having to do the entire project at once.

jjxtra:
True. Probably the nullable reference type makes sense that way. Maybe the asynchronous iterator stuff would be a good first candidate.

PR #472, opened 2019-10-25
C# 8 phase 1

vslee:
- jjxtra began by marking some objects using the nullable (?) reference type signifier


ExchangeSharp/Utility/CryptoUtility.cs
```diff
@@ -1075,37 +1075,39 @@ public static byte[] AesEncryption(byte[] input, byte[] password, byte[] salt)
...
+                AES.Key = key.GetBytes(AES.KeySize / 8);
+                AES.IV = key.GetBytes(AES.BlockSize / 8);
+                AES.Mode = CipherMode.CBC;
+                MemoryStream encrypted = new MemoryStream(input);
+                byte[] saltMatch = new byte[salt.Length];
+                if (encrypted.Read(saltMatch, 0, saltMatch.Length) != salt.Length || !salt.SequenceEqual(saltMatch))
+                {
+                    throw new InvalidOperationException("Invalid salt");
+                }
+                var cs = new CryptoStream(encrypted, AES.CreateDecryptor(), CryptoStreamMode.Read);
```
johnnyasantoss: Maybe here would be good to use `using var` expression.

vslee: Can you add an example of what you mean? Then I can add a commit to this PR. Alternatively, I can merge this PR and you can put it into your current PR or a new one.

johnnyasantoss: I was referring to the same expression in the second example of [this post](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement#example).
Also, I can add it later :)

vslee:
Need @jjxtra to put an approving review before I can merge (I know it's strange for him to be reviewing his own code but since I was the one starting the PR, I can't review my own PR)

jjxtra:
I don't know if I approve of my code to be honest :) We might consider backing out of nullable reference types on a project level and going to a per file level to reduce the warning barrage.

vslee:
Far fewer warnings now (~50 instead of ~500)

vslee:
I think we can merge now, and then whittle the rest of the warnings down with time.

vslee:
progress towards #450

PR #475, opened 2019-11-01
Add buy/sell/cancel to BL3P

johnnyasantoss:
Also, adds some new *shiny* console options for those three options.

<details><summary>Help for the "buy" command</summary>
<p>

```log
ExchangeSharpConsole 0.6.2
Copyright (C) 2019 ExchangeSharpConsole

  --debug            Waits for a debugger to be attached.

  -e, --exchanges    Required. Regex of exchanges to test, null/empty for all.

  --dry-run          (Default: false) Prevents any requests from being
                     performed.

  -k, --key          (Default: keys.bin) Path to key file (generated with the
                     key utility).

  -i, --interval     (Default: 5000) Interval to fetch data in milliseconds.

  -w, --wait         (Default: false) Waits interactively.

  -r, --price        Required. The price to buy or sell at.

  -a, --amount       Required. Amount to buy or sell.

  --stop-price       The price to trigger a stop.
                     This has to be implemented and supported by the chosen
                     exchange.

  --order-type       Required. (Default: Limit) The type of order.
                     Possible values: 'limit', 'market' and 'stop'

  --margin           (Default: false) Whether the order is a margin order. Not
                     all exchanges support margin orders, so this parameter may
                     be ignored.
                     You should verify that your exchange supports margin orders
                     before passing this field as true and expecting it to be a
                     margin order.
                     The best way to determine this in code is to call one of
                     the margin account balance methods and see if it fails.

  --round            (Default: true) Whether the amount should be rounded.
                     Set to false if you know the exact amount, otherwise leave
                     as true so that the exchange does not reject the order due
                     to too many decimal places.

  -s, --symbol       Required. Symbol (currency pair) to be fetched from the
                     exchange.

  --help             Display this help screen.

  --version          Display version information.

```

</p>
</details>


ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
```diff
@@ -138,9 +165,151 @@ Task MessageCallback(IWebSocket _, byte[] msg)
...
+			{
+				request.AddHeader("Rest-Key", PublicApiKey.ToUnsecureString());
+				var signKey = GetSignKey(request, formData);
+				request.AddHeader("Rest-Sign", signKey);
+			}
+		}
+
+		private string GetSignKey(IHttpWebRequest request, string formData)
+		{
+			//TODO: Use csharp8 ranges
```
johnnyasantoss: To use `System.Range` we would have to drop `net472` as a a target framework

ExchangeSharp/API/Exchanges/BL3P/BL3PException.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using System.Runtime.Serialization;
+using ExchangeSharp.API.Exchanges.BL3P.Models;
+
+namespace ExchangeSharp.API.Exchanges.BL3P
+{
+	[Serializable]
```
jjxtra: Should be internal classes

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3PCurrencyFee.cs
```diff
@@ -0,0 +1,8 @@
+namespace ExchangeSharp.API.Exchanges.BL3P.Enums
+{
+	public enum BL3PCurrencyFee : byte
+	{
```
jjxtra: Should be internal

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3POrderType.cs
```diff
@@ -0,0 +1,8 @@
+namespace ExchangeSharp.API.Exchanges.BL3P.Enums
+{
+	public enum BL3POrderType
+	{
```
jjxtra: Should be internal

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3PResponseType.cs
```diff
@@ -0,0 +1,9 @@
+namespace ExchangeSharp.API.Exchanges.BL3P.Enums
+{
+	public enum BL3PResponseType
+	{
```
jjxtra: Should be internal

ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
jjxtra: That's not an option unfortunately right now

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3PResponseType.cs
johnnyasantoss: Done in 9f3bab8.

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3POrderType.cs
johnnyasantoss: Done in 9f3bab8.

ExchangeSharp/API/Exchanges/BL3P/Enums/BL3PCurrencyFee.cs
johnnyasantoss: This can't be made `internal`, because, it's one of the settings of `ExchangeBL3PAPI`. See prop `ExchangeBL3PAPI.DefaultFeeCurrency`.

ExchangeSharp/API/Exchanges/BL3P/BL3PException.cs
johnnyasantoss: Done in 9f3bab8.

ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
johnnyasantoss: We already support `netstandard2.0` which is compatible [¹] [²] with `net461`. We are not using any APIs that are `net472` specific (AFAIK) :thinking:  

 [¹]: https://github.com/dotnet/standard/blob/master/docs/versions.md
 [²]: https://dotnet.microsoft.com/platform/dotnet-standard

jjxtra:
Looks like merge conflict, other than that I'm good to merge in

ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
vslee: Wouldn't that mean that we would have to drop support for netstandard 2.0 as well in order to use System.Range?

johnnyasantoss: That is true. One thing that I said that was wrong is that we need to drop `net472` when actually what I should have said is that there's no need to support `net472` when we already support `netstandard20`. With that clarified, I noticed that those types (`Range` and `Index`) are not available in all runtimes (only `netstandard2.1` and `netcoreapp30`) and in that case we could use a package that ports those specific APIs to `netstandard20` like this one https://github.com/bgrainger/IndexRange

johnnyasantoss:
@jjxtra @vslee Could you check the PR again? :)
Thanks!

ExchangeSharp/API/Exchanges/BL3P/ExchangeBL3PAPI.cs
johnnyasantoss: Maybe we could create an issue out of this conversation :thinking:

vslee: > we could use a package that ports those specific APIs to `netstandard20` like this one https://github.com/bgrainger/IndexRange

Yes that's a good way to keep netstandard20 and start using the new APIs




> Maybe we could create an issue out of this conversation 🤔

Yes feel free

PR #481, opened 2019-11-10
Binance.US: added exchange support

vslee:
- since Binance and Binance.US share the same API, they inherit from BinanceGroupCommon
- this is similar to the refactoring in OKGroup #418
- fixed bug in ExchangeGlobalCurrencyReplacements in BitBank
- added trade stream (websockets) support for Binance DEX
- unfortunately, Binance DEX doesn't share the same API as Binance and Binance.US
- now that a new ExchangeBinanceDEXAPI class is created, others are free to contribute additional implementation details


jjxtra:
Thanks so much for sharing this code. Any change we can make the Binance specific models internal?

PR #492, opened 2019-11-18
change to addChangeLog to false

vslee:
- as discussed in https://github.com/jjxtra/ExchangeSharp/issues/478#issuecomment-554953464

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -151,6 +151,6 @@ jobs:
           title: $(setTag.TAG)
           tag: $(setTag.TAG)
           action: 'edit'
-          addChangeLog: true
+          addChangeLog: false
           assets: '$(System.ArtifactsDirectory)/**'
           assetUploadMode: 'replace'


johnnyasantoss:
> I guess we'll have to wait for the next month upgrade. :)
> Is there any way to automate the release drafter to create a tag every month if there were any changes to the code base?

I believe it is possible to combine Github actions `on: schedule` with the pipeline that builds a new version every month. There's a section on the readme of the project to explain it further. https://github.com/toolmantim/release-drafter#github-actions

vslee:
Currently, the release-drafter is a GH app rather than a GH action. The doc you referenced just refers to how often the release-drafter updates the draft. I think we should keep the release-drafter as is since it is nice to have the notes updated every pull. 
We can create a separate GH action or Azure pipeline step to push a new tag every month.

vslee:
Actually, once https://github.com/toolmantim/release-drafter/issues/141 is resolved, we can have release-drafter tag and publish the releases. A PR resolving is linked from that issue and looks like it's almost to be merged.

PR #498, opened 2019-11-27
NDAX: fix symbol separator bug

vslee:
- NDAX sometimes uses _ as a symbol separater and sometimes it does not
- change the _marketSymbolToInstrumentIdMapping to store the symbol w/o the _
- then search both ways if needed


jjxtra:
Wow whoever made their api was drunk or something. Thanks for doing this...

PR #499, opened 2019-11-29
Binance Jersey: added exchange support

vslee:
- found another Binance...


vslee:
@jjxtra can you take a look?

PR #507, opened 2020-01-13
remove nuget creation from ExchangeSharp.Forms

vslee:
- ExchangeSharp.Forms is just a sample project
- we only publish a nuget from the main ExchangeSharp project


jjxtra:
Sorry I didn't get to this sooner, crazy couple of weeks...

PR #514, opened 2020-01-23
Patch fix place safe market order

### LINKED ISSUE #512: PlaceSafeMarketOrderAsync - Bug
I have found two places that have caused incorrect behaviour in ExchangeAPIExtensions.PlaceSafeMarketOrderAsync():

Line 309 - The "IsBuy" parameter isn't set in the request, so it is always treated as sell-order.

Line 331 - "break" breaks the Switch-statement, not the for-loop. So the loop always runs 20 times (maxTries) and then throws an APIExeption for time out.

jjxtra:
Fixes #512


jjxtra:
Heading to bed, whoever approves please feel free to squash merge it.

PR #515, opened 2020-01-26
fix Base/Quote currency ordering in global symbol conversion

### LINKED ISSUE #500: Making Sense of Base/Quote Currency Ordering for Symbol Strings
Ok, 

So typically a currency pair is expressed as BASE-QUOTE (note ordering)

E.g. 
- EUR-USD is 1.2
- XLM-BTC is 0.0000077

Some exchanges list these reverse, e.g. Bittrex. For these exchanges that list the symbols in reverse, there is a property called MarketSymbolIsReversed which will translate symbols for the exchange properly. However, it does not function as designed.

When translating between global market symbols <->exchange symbols, there is some unexpected behavior based on our assumptions above. See the following evaluations (this is using the Bittrex API where MarketSymbolIsReversed = true):

image

1) gms (global market symbol) above should be "XLM-BTC" because Bittrex symbols are reversed, and XLM is the base currency, so using global standards of BASE-QUOTE, it should be "XLM-BTC" but it does not evaluate as such

image

2) ems (exchange market symbol) should be "BTC-XLM" because again, using global standards of BASE-QUOTE, and Bittrex is reversed, converting "XLM-BTC" global ticker to market ticker should evaluate to "BTC-XLM", which is the Bittrex market ticker
3) ems2 is correctly evaluated to the exchange market symbol for Bittrex composed using the base/quote currency strings as separate inputs
4) gms2, which is the global market symbol of the "BTC-XLM" market on the Bittrex exchange  should be "XLM-BTC", again because XLM is the base currency

The comments in the code are confusing/ambiguous at best and contradictory at worst:

image

Can someone shed some light on this?

image

nh43de:
Fixes #500 

Global symbols list the base currency first (e.g. BTC) and quote/conversion currency
second (e.g. USD). Global symbols are of the form BASE-QUOTE. BASE-QUOTE is read as
1 BASE is worth y QUOTE.

Examples:
	On 1/25/2020,
		- BTC-USD: $8,371; 1 BTC (base) is worth $8,371 USD (quote)
		- ETH-BTC: 0.01934; 1 ETH is worth 0.01934 BTC
		- EUR-USD: 1.2; 1 EUR worth 1.2 USD

Pull Request Changes:
- Fixed issue with GlobalMarketSymbolToExchangeMarketSymbolAsync returning QUOTE-BASE in base ExchangeAPI class
- Fixed tests to use standard BASE-QUOTE format
- Fixed Bitthumb and Kraken exchange APIs as a result of changes
- Added some better details/docs about currency pairs

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/Bithumb/ExchangeBithumbAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/Bithumb/ExchangeBithumbAPI.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
 MIT LICENSE
 
 Copyright 2017 Digital Ruby, LLC - http://www.digitalruby.com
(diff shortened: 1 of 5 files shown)


nh43de:
One more note - this is likely a breaking change because of how global symbols were treated in some cases.

jjxtra:
Anyone have any objections here?

vslee:
Will defer to you on this one

jjxtra:
@vslee Can we rev to a 0.7.0 version as this is a breaking change?

vslee:
Yes, I'll set the next nuget version to be 0.7.0.

jjxtra:
Thank you!

PR #531, opened 2020-04-02
Digifinex: API improvements

szmcdull:
add more servers and find the fastest one at runtime
solve 10008 time error
raise rate limit to 240 times per minute
candle add `limit` parameter support

ExchangeOKExAPI.cs not changed. Don't know why it is listed.


vslee:
Can you rebase so the commits don't include a bunch of old ones? That's the reason that ExchangeOKExAPI.cs is showing up here.

szmcdull:
@vslee I tried it before but it was very difficult to rebase because the directory structure has changed

vslee:
How do you normally work w/ git. [Git Extensions](http://gitextensions.github.io/)?

szmcdull:
@vslee Git Bash & TortoiseGit

vslee:
Okay, here's how to prevent this from happening in the future. Before making any changes to the code, fetch/pull the latest code from the jjxtra master. Then, reset your local to the latest commit (and discard any local changes). After that, you can make new changes to the code. When you're done with that, you have to force push it up to your github. Then when you make a PR, it will show up as just one commit.

szmcdull:
@vslee Could you please provide a step by step command instruction. I don't want to make a mistake next time.

szmcdull:
What if the upstream changed while I'm making changes to my branch?

vslee:
I don't use either Git Bash or TortoiseGit, so I can't give you instructions on how to use those. I do use Git Extensions though, so if you want to use that, I can show you how.

> What if the upstream is modified while I'm making changes to my branch?

If the upstream changes are in different files than the ones you're working on, then you can just ignor e them and not fetch the new changes. Then force push your commit up to your branch. When you create a pull request, it won't affect the upstream changes. 
If the upstream changes do affect the same files, then you will need to stash your local changes first. After that, you can fetch the new upstream changes and reset your local. Then apply the stash.

vslee:
I should add, don't worry about "making a mistake next time". If it doesn't work for your next PR, we can still figure it out.

PR #565, opened 2020-12-17
Convert exchange currencies format to global currencies format in the ExchangeAPI class

borseno:
# Problem
Some exchanges such as Kraken and Binance etc. have weird currencies namings and these weird namings are considered to be of "exchange" format and the most popular namings are considered to be of "global" format. And in the GetAmounts etc. methods the keys returned (strings) are expected to be of global format, not exchange. But the exchanges return exchange format

# Solution
I added in the ExchangeAPI class mapping from exchange to global currency format after calling OnGetCurrencies etc methods. Also, some exchanges require initialization to be done (Kraken for example) so that it can come up with mapping tables, so I decided to add InitializeAsync method

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/Kraken/ExchangeKrakenAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/Kraken/ExchangeKrakenAPI.cs
@@ -44,11 +44,6 @@ namespace ExchangeSharp
 
         static ExchangeKrakenAPI()
         {
-            ExchangeGlobalCurrencyReplacements[typeof(ExchangeKrakenAPI)] = new KeyValuePair<string, string>[]
-            {
-                new KeyValuePair<string, string>("XBT", "BTC"),
-                new KeyValuePair<string, string>("XDG", "DOGE")
-            };
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


vslee:
@jjxtra worked on the global currency mapping previously so I'll defer this to him

pomeara:
Great pickup - I have had similar thoughs previously however did not get around to actually implementing a fix - great stuff @BadgerTheAdger 

@jjxtra have you had a chance to check this out - would be great to have this feature in the repository

jjxtra (changes requested):
I don't have a strong opinion on global currencies. If someone has an idea and wants to refactor, have at it :)

I like the idea of InitializeAsync, however I would like to see it called only once or on an interval, and be protected. In this pr, there are cases where it will be called multiple times, which could impact performance.

I think we should remove the GetExchangeAPIAsync method as it is confusing, there are multiple methods to get an exchange. We can solve the initialize part by doing the following:

Use a protected OnInitializeAsync in the ExchangeAPI class (put it right below OnDispose) and impement a try/catch block there with some sort of lock (SemaphoreSlim is great for this sort of thing for async methods), preventing multiple initialize from stacking, etc. Once it succeeds successfully, set a timestamp so the method is only entered every hour. Most of the public methods in ExchangeAPI would be great places to add the await OnInitializeAsync() calls, that way derived exchanges don't know or care about it, unless they override OnInitializeAsync themselves.

Whenever you see a `new SynchronizationContextRemover`, that would be a great place to call the OnInitializeAsync right below it.

jjxtra:
Closed by latest commit to master 9cf6379dda9d97d9f0b9043946e7c371491c0893

Issue #584, opened 2021-05-01
OrderType is Market but it places a limit buy order?


Diogovski:
Hi, i am facing a problem. I am trying to place a market order and after that show the order details on the console! But idk why the function is placing a limit buy order, and because of that i am not able to pull some details about the order such as amount filled, price and fees. I'll attach the code! Am i doing something wrong?Is this a bug?

ExchangeTicker ticker = kucoin.GetTickerAsync("XLM-USDT").Sync();
            ExchangeOrderResult result = kucoin.PlaceOrderAsync(new ExchangeOrderRequest
            {
                
                Amount = 1,
                OrderType = OrderType.Market,
                IsBuy = true,
                Price = ticker.Ask,
                MarketSymbol = "XLM-USDT"
            }).Sync();
            System.Threading.Thread.Sleep(500);
            var orderdetails = kucoin.GetOrderDetailsAsync(result.OrderId).Sync();
             Console.WriteLine("Average Price: " + orderdetails.AveragePrice);
             Console.WriteLine("Amount Filled: " + orderdetails.AmountFilled);
             Console.WriteLine("Fee Currency: " + orderdetails.FeesCurrency + "\n" + "Total Fees: " + orderdetails.Fees);
             Console.WriteLine("Order Date: " + orderdetails.OrderDate);

vslee:
There may be a bug within the ExchangeSharp code. If you want to investigate and come up with a fix, we will be happy to incorporate it.

Diogovski:
how do i  edit the l ExchangeSharp code?

vslee:
To create a pull request, you have to fork the repository first. Then on your fork, create a new branch and commit your changes there. Then, go back to github and view your fork, there will be a button to create a pull request. 

What code editor do you use?

Diogovski:
visual studio

vslee:
There is a extension for VS that can help you https://visualstudio.github.com/

vslee:
Here is a walkthrough  https://docs.microsoft.com/en-us/learn/modules/visual-studio-github-push/

Diogovski:
thank you

onurselcuk:
When you add "price" field into request, it will be converte to limit order by kucoin. Can you test without "price"

PR #591, opened 2021-05-06
KuCoin market order bugfix

onurselcuk:
#584


vslee (approved):
Wow, you're on a roll!

PR #593, opened 2021-05-15
Add new proeprty to ExchangeTicker that contains response JSON. This …

jacobdeboer33:
…to get additional data (if needed and provided)

### DIFF
--- a/src/ExchangeSharp/Model/ExchangeTicker.cs
+++ b/src/ExchangeSharp/Model/ExchangeTicker.cs
@@ -53,7 +53,7 @@ namespace ExchangeSharp
 		public decimal Last { get; set; }
 
 		/// <summary>
-		/// This property contains the content of the complete api request results. This may contain additional information
+		/// This property contains the content of the complete api response. This may contain additional information
 		/// </summary>
 		public JToken ApiResponse { get; set; }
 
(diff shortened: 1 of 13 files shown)


vslee:
This PR is on the larger side, so I will defer to @jjxtra.

jdx-john:
I took a look out of interest but at least on GitHub's website, the diff is really screwed up?

src/ExchangeSharp/Model/ExchangeTicker.cs
```diff
@@ -21,109 +21,114 @@ MIT LICENSE
...
+
+		/// <summary>
+		/// The last trade purchase price
+		/// </summary>
+		public decimal Last { get; set; }
+
+		/// <summary>
+		/// This property contains the content of the complete api request results. This may contain additional information
+		/// </summary>
+		public string ApiResultString { get; set; }
```
vslee, line 57: Instead of storing this as a large string, can you just store the JToken? That way, users don't have to incur a large `ToStringInvariant()` for every ticker response.

vslee (changes requested):

src/ExchangeSharp/Model/ExchangeTicker.cs
jacobdeboer33: I think you're right. I've updated the property name and type.

vslee (approved):

Issue #595, opened 2021-06-01
Candles in websockets?


jdx-john:
I notice that Kraken's WS API provides `ohlc` but as far as I can see ExchangeSharp only exposes the REST version.
Is this a deliberate design or perhaps it's not seen as useful?
I wouldn't be surprised if there's already a utility class which will generate candles from the raw trades feed... I keep starting to write things then stumbling across existing implementations!

vslee:
Streaming tickers is implemented in ExchangeSharp for many exchanges but not Kraken. If you'd like you submit an implementation as a Pull Request, we'd be happy to merge it!

jdx-john:
@vslee I'm not referring to tickers, but to OHLC candles.

I thought that WS functionality for Kraken _is_ supported, e.g. `OnGetTickersWebSocketAsync` / `OnGetTradesWebSocketAsync` in `ExchangeKrakenAPI`?

But my question is that in `IExchangeAPI`, there is no signature defined at all for streaming OHLC data. Perhaps KRaken is the only exchange who added this (https://docs.kraken.com/websockets/#message-ohlc) or maybe everyone builds their own candles from raw trades data so it's not considered important?

So to submit a PR I would need to add a new API method on the interface and presumably dummy versions for every other exchange, as well as doing the implementation?

vslee:
I see what you're saying, ExchangeSharp only had REST support for Candles, but not Websocket support. I just made a PR which adds Websocket support for candles. After my PR is merged, you can then add OHLC support to Kraken by overriding the `OnGetCandlesWebSocketAsync()` method in `ExchangeKrakenAPI.cs`.

PR #596, opened 2021-06-01
Add Candles (OHLC) websocket support in ExchangeAPI

vslee:
- part of #595

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
@@ -178,7 +178,7 @@ namespace ExchangeSharp
 			throw new NotImplementedException();
 		protected virtual Task<ExchangeCloseMarginPositionResult> OnCloseMarginPositionAsync(string marketSymbol) =>
 			throw new NotImplementedException();
-		protected virtual Task<IWebSocket> OnGetCandlesWebSocketAsync(Action<IReadOnlyCollection<MarketCandle>> callback, params string[] marketSymbols) =>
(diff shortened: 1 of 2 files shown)


jdx-john:
I don't think I'm placed to review but this seemed a good place to ask a question, as I am interested to provide a Kraken implementation.

Why are the method signatures not consistent?

```
OnGetCandlesWebSocketAsync(Action<IReadOnlyCollection<MarketCandle>> callback, params string[] marketSymbols)
OnGetTickersWebSocketAsync(Action<IReadOnlyCollection<KeyValuePair<string, ExchangeTicker>>> tickers, params string[] marketSymbols)
OnGetTradesWebSocketAsync(Func<KeyValuePair<string, ExchangeTrade>, Task> callback, params string[] marketSymbols) 

```

1. Trades passes a `Func` rather than `Action` but looking at the console they are used exactly the same
2. Why do we have return type of `<string, ...>` on Tickers & Trades but not Candles? I _think_ the string holds the pair-name, but isn't that returned inside the objects anyway?

Doesn't stop me working on a Kraken implementation in the meantime locally.

vslee:
There was not an API design process in the beginning. Instead, it has evolved over the ears. Various contributors have added bits and pieces. That is why there are inconsistencies. Perhaps one day someone will want to go through and clean up the API. In the meantime, we can do what makes the most sense for any new additions. 
You're right in that returning a `string` for the symbol is superfluous, since the symbol name already exists inside `MarketCandle`, so that's why I have left it off. The reason that Trades passes a `Func` instead of `Action` is that this allows the callback itself to be an `async` method. Come to think of it, this would be useful for the candles callback so I will push a commit that changes this. Eventually, the rest of the methods can be updated to support this as well, but for now we can just update this one.

jdx-john:
Thanks @vslee it's good to know I wasn't missing anything major with it being a new codebase to me. I hadn't thought of the `async` callback detail, though. That's a good point.

jjxtra (approved):

PR #611, opened 2021-07-13
Fixes #610 - Use TryAddWithoutValidation() when setting HTTP headers. 

### LINKED ISSUE #610: PlaceOrderAsync throwing exception on Poloniex after changing to System.Net.Http.HttpClient.
Exception is thrown after calling PlaceOrderAsync method on Poloniex.
I've also tested same method on Binance and it worked fine.
It happened after this commit. [babff917815d52742491bb12ae77b869ac7748ea]

Exception details: 
InvalidOperationException
ExchangeSharp.APIRequestMaker.InternalHttpWebRequest.AddHeader(String header, String value) in D:\ExchangeSharp-master\src\ExchangeSharp\API\Common\APIRequestMaker.cs:line 93
Misused header name. Make sure request headers are used with HttpRequestMessage, response headers with HttpResponseMessage, and content headers with HttpContent objects.

The problematic header seems to be `content-length`

Solution:
Changing `Request.Headers.Add(header, value);` to `Request.Headers.TryAddWithoutValidation(header, value);`

EmergentCybernetics:
* Use `TryAddWithoutValidation()` when setting HTTP headers. 
* Switch `APIRequestMaker.cs` to tab indention.
* Allow override of `ConnectWebSocketAsync()` and friends to prep for Bittrex v3 WebSocket support.

### DIFF
--- a/src/ExchangeSharp/API/Common/APIRequestMaker.cs
+++ b/src/ExchangeSharp/API/Common/APIRequestMaker.cs
@@ -21,110 +21,110 @@ using System.Threading.Tasks;
 
 namespace ExchangeSharp
 {
-    /// <summary>
-    /// Handles all the logic for making API calls.
-    /// </summary>
-    /// <seealso cref="ExchangeSharp.IAPIRequestMaker" />
-    public sealed class APIRequestMaker : IAPIRequestMaker
-    {
-        private readonly IAPIRequestHandler api;
+	/// <summary>
+	/// Handles all the logic for making API calls.
+	/// </summary>
+	/// <seealso cref="ExchangeSharp.IAPIRequestMaker" />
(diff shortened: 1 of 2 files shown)


src/ExchangeSharp/API/Common/APIRequestMaker.cs
```diff
@@ -21,110 +21,110 @@ MIT LICENSE
...
+
+			public void AddHeader(string header, string value)
+			{
+				switch (header.ToLowerInvariant())
+				{
+					case "content-type":
+						contentType = value;
+						break;
+					default:
+						Request.Headers.TryAddWithoutValidation(header, value);
```
EmergentCybernetics, line 84: Only actual change is here for #610. The rest is linting.

vslee (approved):

PR #615, opened 2021-07-22
Make market orders possible by removing price requirement

raymens:
This change was required to make Marketorders possible on CoinbasePro.
Not sure what the other echanges allow and or require so I've kept it simple by just assuming the price is given.

I suppose it's also possible to add checks and fail in a nicer way, what's your take on it?


vslee (changes requested):
Yes, it would be great to add checks and throw an ArgumentNullException in  the two cases, instead of having a more cryptic exception when accessing the value.

vslee:
Superseded by #620

Issue #617, opened 2021-07-22
KrakenExchange.OnGetTickersAsync() throws Null.


jdx-john:
https://github.com/jjxtra/ExchangeSharp/blob/4ac4ece56447cac17226811e8f9a06e5d07278f9/src/ExchangeSharp/API/Exchanges/Kraken/ExchangeKrakenAPI.cs#L564

This fails to check `symbol == null` which is a trivial fix but I'm not sure if this is masking a deeper problem. When I test, it fails on `ETCETH`. `GetMarketSymbolsAsync` returns ETC/ETH but `GetMarketSybolsMetadataAsync` has:
![image](https://user-images.githubusercontent.com/1654627/126710598-06497a18-c61b-4239-a324-d396e5bfc259.png)

It seems quite an easy match; not sure why it's doing this raw string manipulation when alt fields contain the data. But I'm too tired to try a fix right now without breaking something :)

vslee:
Since that was implemented long ago, I don't think we'll ever figure out why. Some ideas I can think of are:

- The alt field may not have existed previously
- The implementer didn't have a good view of the all of the properties and did not know about the existence of the alt field
- The alt field may not be consistent across all symbols
- some other reason not immediately apparent

Regardless, I think it would be fine to try changing to using the alt field instead

jdx-john:
@vslee I tested changing to:

```
					var symbol = symbols.FirstOrDefault(a => a.AltMarketSymbol.Equals(marketSymbol));
					if (symbol == null)
						continue;
```

I now get a ticker for all 365 expected markets - that `continue` is never hit. I haven't checked every ticker is returning the _correct_ value but the ones I looked at were.

It does make me nervous to silently ignore errors though, but that seems more an architecture question. I have wondered about something simple like adding `IExchangeAPI.LastError` which any API method can populate e.g "we expected 354 tickers but only found 349". Kind of crude but also open-ended. Someone was suggesting similarly allowing raw JSON responses to be retrieved a fe weeks back, I think?

vslee:
> It does make me nervous to silently ignore errors though, but that seems more an architecture question. 

Ideally, the API would have been designed so all methods returned an enclosing object that had a `Data` property as well as an `Error` property. It's too late to change everything now though. 

> I have wondered about something simple like adding `IExchangeAPI.LastError` which any API method can populate e.g "we expected 354 tickers but only found 349". Kind of crude but also open-ended. 

[...]

PR #618, opened 2021-07-24
Better symbol name translation and error detection in  KrakenExchange.OnGetTickersAsync()

jdx-john:
See #617 - we use `AltMarketSymbol` now, and also log any non-translated symbols.

I am unclear why my PR includes merge commits and so on, the only changed file is as expected but it seems messy. Apologies if I have done something wrong, all I did:
- make changes in my fork
- commit, push in my fork
- from GitHub, fetch upstream to get my fork up to date
- from GitHub, make a pull-request 

I'm not used to working with forks, is this incorrect?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -457,3 +457,6 @@ launchSettings.json
 dist/
 data/**
 !data/.gitkeep
+
+## Mac specific
+.DS_Store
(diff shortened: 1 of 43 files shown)


vslee (approved):
In order to avoid all of those extra commits, you can start by fetching upstream as your first step. Once your local branch is up to date with the master, then that's when you start making your new changes. As long as the master doesn't change after that, then you won't need to fetch the upstream again before making your PR.

PR #620, opened 2021-08-01
Fixed market order bug using Order.Price

vslee:
Continuation of #615 with requested changes.
Original PR comment by @raymens
> This change was required to make Marketorders possible on CoinbasePro.
> Not sure what the other echanges allow and or require so I've kept it simple by just assuming the price is given.
> 
> I suppose it's also possible to add checks and fail in a nicer way, what's your take on it?


raymens:
Thanks a lot, I had it on my list of to-dos. But I appreciate this!

jjxtra (approved):

Issue #624, opened 2021-08-11
Is there a sample OrderBook implementation?


jdx-john:
I'm putting together a simple implementation that uses the `SortedDictionary` of Asks and Bids for Websockets (I guess it would work for REST too). I think this is consistent across all exchanges so I can contribute it but it seems like something that might already exist.
Kraken provides Checksum functionality to test your book hasn't gotten out of sync, not sure if others do this but ES doesn't currently expose this at all so I'd at least ad support in `ExchangeOrderBook`. Other than that it's basically just merging dictionaries.

vslee:
Nothing outside of what is in this repo already. But if you do end up creating one, we'd be happy to merge it.

jdx-john:
I was wondering if there was something lurking in the repo... I keep finding things! Very happy to share it if I can avoid making it specific to my application.

PR #626, opened 2021-08-12
Fix Kraken WS orderbook updates and add checksum support

jdx-john:
The existing Kraken implementation for order-book updates is incorrect. It assumes each update will include either one "a" or one "b" block and that the message contains exactly 5 elements.
In most cases this is the case however the API can and sometimes does send both an "a" and "b" block in the same message. I have verified this myself with a live feed (examples are is also shown in the API docs) and in this case the behavior is faulty. The market symbol is read wrongly and the whole message will therefore get missed leading to order-book mismatch locally.

My change follows the spec more accurately and also adds support for Kraken's checksum property (https://docs.kraken.com/websockets/#book-checksum). I do not know if other exchanges provide this but it is _vital_ to ensure your local order book is correct in case of missed messages. I have implemented the simplest possible API change for `ExchangeOrderBook.Checksum` which could be used by other exchange implementations if desired.

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/Kraken/ExchangeKrakenAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/Kraken/ExchangeKrakenAPI.cs
@@ -1142,6 +1142,15 @@ namespace ExchangeSharp
 			return marketSymbolList;
 		}
 
+		/// <summary>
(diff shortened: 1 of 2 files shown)


jdx-john:
Assuming this gets accepted I'll add an implementation of the Kraken checksum as a utility method on `ExchangeKrakenApi`, linked to #624

vslee (approved):

PR #647, opened 2021-09-16
Added ability to GetOrderDetailsAsync() by ClientOrderId

vslee:
- many, but not all, exchanges support this


jjxtra (changes requested):
Thanks for the pr. I request that you move this new optional parameter to the end of the method so that we don't break everyone who updates to this new version.

vslee:
Done

jjxtra (approved):

PR #651, opened 2021-09-19
Standardize isPostOnly

vslee:
- a.k.a. MakerOnly or BookOrCancel or PendingOrCancelled
- most exchanges support this, but vary in the parameters needed to specify this


jjxtra (approved):

vslee:
@brigb123 and @yatendra, note starting with v0.9.0, `IsPostOnly` can be set though `ExchangeOrderRequest`. If it is left `null`, then it will revert to the exchange default, likely `false` for Coinbase.

Issue #634, opened 2021-08-26
FTX API Implementation


Plumbly:
Hi all,

Is anyone planning, or currently in the process of, integrating [FTX](https://ftx.com/)? If not, I will look to implement it myself and make a pull request. I just wanted to make sure I don't spend time on it if someone is already on it.

Plumbly:
Just a quick update that I should be ready to make a pull request for this exchange implementation by the end of the week.

bezysoftware:
Are you planning to add withdrawals (+withdrawal fees)? I think it shouldn't be too complicated based on their docs

Plumbly:
I did plan to do it at some point, as I will need it for the project I work on, but didn't need it right now. However, you are right, it doesn't look like too much work so, will try find some time to add it.

bezysoftware:
I think there's also an issue with the namespace. It needs to be in "ExchangeSharp" namespace for the lookup by name to work: https://github.com/jjxtra/ExchangeSharp/blob/master/src/ExchangeSharp/API/Exchanges/_Base/ExchangeName.cs#L50

vslee:
> I think there's also an issue with the namespace. It needs to be in "ExchangeSharp" namespace for the lookup by name to work: https://github.com/jjxtra/ExchangeSharp/blob/master/src/ExchangeSharp/API/Exchanges/_Base/ExchangeName.cs#L50

Yes, we'll have to fix that before the next NuGet

Plumbly:
I have now fixed the namespace of the FTX exchange.

PR #662, opened 2021-09-27
FTX Exchange Implemented

Plumbly:
This should be the FTX exchange now implemented as per #634.


Plumbly:
Apologies, just realise I included a commit that I did not want in then.  I will close and reopen.

PR #667, opened 2021-09-30
Added Expired and PendingOpen to ExchangeAPIOrderResult

vslee:
This PR contains the following breaking changes to the ExchangeAPIOrderResult enum:
- `PendingOpen` is an important state that many exchanges have, which is separate from `Open`
- Renamed `Pending` state to `Open`
- Renamed `Error` state to `Rejected`, since this is how most exchanges refer to it
- Added `Expired` state, which is separate from `Rejected`

also contains the following changes
- added stricter checks to FTX` OnGetOrderDetailsAsync()` and `OnPlaceOrderAsync` `isPostOnly`
- improved FTX `ToExchangeAPIOrderResult()`, renamed to `FTXExtensions` to reduce possibility of collision in the future

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/Aquanow/ExchangeAquanowAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/Aquanow/ExchangeAquanowAPI.cs
@@ -189,7 +189,7 @@ namespace ExchangeSharp
 					break;
 
 				default:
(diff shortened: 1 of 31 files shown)


jjxtra (approved):

jjxtra:
Great stuff ;)

PR #668, opened 2021-09-30
Async ExchangeAPI init

### LINKED ISSUE #666: InitializeAsync is called with .Sync() which blocks the calling UI thread
I just noticed that for Kraken calling `ExchangeAPI.GetExchangeAPI` blocks the entire UI thread and freezes the application. After some digging I think that's caused by this:

https://github.com/jjxtra/ExchangeSharp/blob/master/src/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs#L74 which looks like a ticking time bomb :) 
I think the api should change to be Async, which would allow to await the initialize method, instead of calling `.Sync()`.

bezysoftware:
What I though would be a quick and easy change escalated rather quickly :) The existing sync methods are marked as obsolete with a hint to use the async ones.
The semaphore in `GetExchangeAPIAsync` is used to make sure initialization is called only once 

resolves #666 (😈)


src/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
```diff
@@ -398,75 +400,117 @@ private static IExchangeAPI CreateExchangeAPI(Type? type)
...
 		/// </summary>
 		/// <typeparam name="T">Type of exchange to get</typeparam>
 		/// <returns>Exchange API or null if not found</returns>
+		[Obsolete("Use the async version")]
 		public static IExchangeAPI GetExchangeAPI<T>() where T : ExchangeAPI
+		{
+			return GetExchangeAPIAsync<T>().Result;
+		}
+
+		public static Task<IExchangeAPI> GetExchangeAPIAsync<T>() where T : ExchangeAPI
```
vslee, line 441: Can you update the example in README.md to this new method?

vslee (changes requested):

src/ExchangeSharp/API/Exchanges/_Base/ExchangeAPI.cs
bezysoftware: Done

vslee (approved):
This is technically not a "breaking change", since you are using the `Obsolete` attribute, but I will mark it with the "breaking change" tag to alert ppl to migrate.

PR #690, opened 2021-11-26
Implemented OnGetCandlesWebSocketAsync for BinanceGroup

egehlen:
Implemented websocket Kline/Candlestick Streams following Binance docs:
[https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#klinecandlestick-streams](url)

The data is received each 2 seconds. I know that MiniTicker method (right below this method in the docs) is faster (it updates each 1s) but there's no aggregation by period. That's why I believe this method is more suitable for this case.

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/BinanceGroup/BinanceGroupCommon.cs
+++ b/src/ExchangeSharp/API/Exchanges/BinanceGroup/BinanceGroupCommon.cs
@@ -1135,6 +1135,36 @@ namespace ExchangeSharp.BinanceGroup
 			});
 		}
 
+		protected override async Task<IWebSocket> OnGetCandlesWebSocketAsync(Func<MarketCandle, Task> callbackAsync, int periodSeconds, params string[] marketSymbols)
+		{
+			//Candlesticks are received each 2s. Miniticker method is faster (1s), but you can't aggregate by period.
+			//https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#klinecandlestick-streams
+
+			if (marketSymbols == null || marketSymbols.Length == 0)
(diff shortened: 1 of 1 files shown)


BZ-CO:
Why you are subscribing only for the first symbol in the array?
What if marketSymbols array is empty?

egehlen:
@BZ-CO this method in binance websocket allows only one symbol per connection (see [docs](https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#klinecandlestick-streams)), if user needs to subscribe to more symbols, new connections should be made. Regarding your second question, in my commit, line 1143 is already checking for marketSymbols null or length zero.

BZ-CO:
I've opened a new PR which addresses both of my concerns.
Now you can subscribe for more than one pair on a single connection and to provide empty marketSymbols as well.

Please close this one.

src/ExchangeSharp/API/Exchanges/BinanceGroup/BinanceGroupCommon.cs
```diff
@@ -1135,6 +1135,36 @@ protected override async Task<IWebSocket> OnUserDataWebSocketAsync(Action<object
 			});
 		}
 
+		protected override async Task<IWebSocket> OnGetCandlesWebSocketAsync(Func<MarketCandle, Task> callbackAsync, int periodSeconds, params string[] marketSymbols)
+		{
+			//Candlesticks are received each 2s. Miniticker method is faster (1s), but you can't aggregate by period.
+			//https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#klinecandlestick-streams
+
```
BZ-CO, line 1142: foo

PR #708, opened 2021-12-16
Update README.md description

### LINKED ISSUE #697: Request to add an extra Exchange (I included the code and sample for you) [For testing ExchangeSharp without funds]
Hi,

I have no idea who is responsible for creating a new version of ExchangeSharp, but I found an important (for me) exchange missing.
Perhaps when a new version comes out someone is willing to add my code. (permission etc. is hereby granted to the whole world).

I am new at ExchangeSharp and I am simply learning how to use it. Somewhere on github it says ExchangeSharp comes with extensive documentation. While this may have been true in the past, I cannot find a real manual or document that explains to newbies like me how to use this.

As you may imagine I do not want to test several functions (as to what exectly does this function do) on a live account of an exchange.

Binance has the option of a test account.

I expected that for instance GetAmountsAsync would give me the amounts per Coin in possession, but I wasn't sure.
And because I was able to test this on the test version of the Binance API I now am sure.

Adding the Binance test exchange is very simple. I downloaded the ExchangeSharp source and in src\ExchangeSharp\API\BinanceGroup I added a new class (ExchangeBinanceTestAPI.cs):

using ExchangeSharp.BinanceGroup;

namespace ExchangeSharp
{
	public sealed class ExchangeBinanceTestAPI : BinanceGroupCommon
	{
		public override string BaseUrl { get; set; } = "https://testnet.binance.vision";
		public override string BaseUrlWebSocket { get; set; } = "wss://stream.binance.com:9443";
	}

	public partial class ExchangeName { public const string BinanceTest = "BinanceTest"; }
}

Basically only the BaseUrl is different from ExchangeBinanceAPI

For people that want to use this: Do not forget to request API keys for the Test API. These are different keys from they real Binance API.

When you request these test API keys you get some funds (of course not real) to play around with in your Test API.

Ideal for getting to know for instance how to work with ExchangeSharp!
The test API has fewer crypto coins than the real API and it's value lags behind from the real value. But overall for testing it is ideal.

I have my own application in which I want to use ExchangeSharp and I used Nuget packet manager: Install-Package DigitalRuby.ExchangeSharp -Version 0.9.1 to be able to use this.

But as mentioned I changed the code in the source. So what I needed to do in my program was to delete the reference to ExchangeSharp.dll and include the new ExchangeSharp.dll from my source.

The following code now works as expected:
        static private string BinanceApiKey = "EncryptedBinanceApiKey";
        static private string BinanceApiPrivateKey = "EncryptedBinanceApiPrivateKey";
        static private string BinanceTestApiKey = "EncryptedBinanceTestApiKey";
        static private string BinanceTestApiPrivateKey = "EncryptedBinanceTestApiPrivateKey";

        private void FormBot_Load(object sender, EventArgs e)
        {//Utils.Decrypt string decrypts the keys to there true value
            BinanceApiKey = Utils.DecryptString(BinanceApiKey);
            BinanceApiPrivateKey = Utils.DecryptString(BinanceApiPrivateKey);
            BinanceTestApiKey = Utils.DecryptString(BinanceTestApiKey);
            BinanceTestApiPrivateKey = Utils.DecryptString(BinanceTestApiPrivateKey);
        }

        private static async Task<ExchangeBinanceAPI> CreateBinanceAPI(string response = null)
        {//This would be for real 
            var api = (await ExchangeAPI.GetExchangeAPIAsync(ExchangeName.Binance) as ExchangeBinanceAPI);
            api.LoadAPIKeysUnsecure(BinanceApiKey, BinanceApiPrivateKey);
            return api;
        }

        private static async Task<ExchangeBinanceTestAPI> CreateBinanceTestAPI(string response = null)
        {//this is for testing
            var api = (await ExchangeAPI.GetExchangeAPIAsync(ExchangeName.BinanceTest) as ExchangeBinanceTestAPI);
            api.LoadAPIKeysUnsecure(BinanceTestApiKey, BinanceTestApiPrivateKey);
            return api;
        }

        private async void FetchTicker()
        {//Fetching all tickers is way faster than fetching something like 10 tickers, so I fetch all and filter what I need
	 //Putting them in a listbox is just to compare the output with the original Binance API
            listBox1.Items.Clear();
            this.UseWaitCursor = true;
            ExchangeTicker ticker = new ExchangeTicker();

            try
            {
                var api = CreateBinanceTestAPI();
                var tickers = await api.Result.GetTickersAsync();

                for (int i = 0; i < WhiteList.BaseInfoList.Count(); i++)//WhiteList.BaseInfoList contains a list of crypto coins I am interested in
                {
                    if (WhiteList.BaseInfoList[i].CoinMarketName != WhiteList.BaseInfoList[i].CoinName)
                    {
                        for (int j = 0; j < tickers.Count(); j++)
                        {
                            if (tickers.ElementAt(j).Key == WhiteList.BaseInfoList[i].CoinMarketName)
                            {
                                listBox1.Items.Add(tickers.ElementAt(j).Key + " | " + tickers.ElementAt(j).Value);
                                break;
                            }
                        }
                    }
                }
            }
            catch (Exception ex)
            {
                //textTickersResult.Text = ex.ToString();
            }
            finally
            {
                Invoke(new Action(() => this.UseWaitCursor = false));
            }
        }

        private async void GetCurrentCoinsInPossession()
        {
            var api = CreateBinanceTestAPI();
            var amounts = await api.Result.GetAmountsAsync();
            for (int i = 0; i < amounts.Count; i++)
            {
                string coinName = amounts.ElementAt(i).Key;
                string quantityInPossesion = amounts.ElementAt(i).Value.ToString();
            }
        }

As mentioned it would be nice if this could be included in a newer version of ExchangeSharp.

Also I hope some other newbies can use this to learn how to work with ExchangeSharp.

Kind regards,

Clemens Linders

vslee:
- remove "well documented"
- fixes #697

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Feel free to visit the discord channel at https://discord.gg/sHCUHH3 and chat wi
 ### Features
 
 - Many exchanges supported with public, private and web socket API
-- Easy to use and well documented code and API
+- Easy to use code and API
 - Optional global market symbol normalization, since each exchange has their own way of doing market symbols
 - Runs anywhere .NET runs. (Windows, Mac, Linux, Containers, Serverless, iOS, Android, [etc.](https://docs.microsoft.com/en-us/dotnet/core/about))
 - Can be used from [many different C# platforms](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard2.0.md#platform-support)


jjxtra (approved):

jjxtra:
The code is the documentation lol.

vslee:
Our XML docs aren't bad. But we just don't have the Getting Started writeups and Examples that other projects have. For example: https://joshclose.github.io/CsvHelper/

Issue #713, opened 2022-01-08
OrderBook Websocket: wrong break condition ? 


chabar:
Hi all,

First things: thumbs up for this library !

I'm trying to listen to OKEx Order book market data through websocket api. 

Since "Moving to OKEx V5 WebSocket API (#685)", I noticed that I have wrong order book after a period of time (best bid > best ask). 

It's easily reproductible by launching : exchange-sharp.exe ws-orderbook -e OKEx -s BTC-USD-SWAP
After a while bid and ask are inverted and can remain this way for a long time. 

I strongly suspect ParseOrderBookFromJTokenArrays method:

```
				if (book.Asks.Count == maxCount)
				{
					break;
				}
```

By using the default setting maxCount = 20, there is an inconsistency between the request which is 'books-l2-tbt' and maxCount.
From https://www.okex.com/docs-v5/en/#websocket-api-public-channel-price-limit-channel:
	_"books-l2-tbt: 400 depth levels will be pushed in the initial full snapshot. Incremental data will be pushed every 10 ms when there is change in order book."_

The break condition hides updates with 0 quantity that should remove levels. 
Those levels remain in the order book and will later create false bid or ask on top of the book depending on the market direction.
 
An easy fix/workaround is to use a much higher maxCount for this market but I'm wondering why this break condition exists in the first place.

Though very unlikely, nothing prevents the exchange to replace all price levels in one update.
You would received all previous prices with 0 quantity and all new prices, that's order book depth x 2.

I check the updates received from OKEx over a period of 15 min. This is the distribution of update count (bid + ask):
![image](https://user-images.githubusercontent.com/5296273/148651574-8a26274a-9dac-4030-8f23-d51431608735.png)

(every occurence above 40 can create a issue in case maxCount is 40)

I would propose to remove the condition at parsing level. 
I don't know all the use cases of the library so I wondering what the community think about this.

Regards,

Charles

BZ-CO:
Hi.

I also think that trimming should be done at `GetFullOrderBookWebSocketAsync` rather than how it's now on the incoming data.
Order book store is in `GetFullOrderBookWebSocketAsync,` so it should be responsible for the max count too.

I've made a sketch and will provide PR tomorrow.

Would be happy to hear what other thinks as well.

jjxtra:
@BZ-CO Good plan

jjxtra:
Resolved https://github.com/jjxtra/ExchangeSharp/pull/714

PR #714, opened 2022-01-10
fix GetFullOrderBookWebSocketAsync data inconsistency

BZ-CO:
Proposal for removing `maxCount` parameter from `ParseOrderBookFromJTokenArrays` and `ParseOrderBookFromJTokenDictionaries`.
Currently we are trimming the incoming order book data before processing it, which seems wrong and causes order book inconsistency as mentioned in #713 .
Instead I propose moving `maxCount` trimming logic to `GetFullOrderBookWebSocketAsync`, because that's the method responsible for order book data store as well.


jjxtra (approved):
Thanks for this refactor, much cleaner

PR #721, opened 2022-01-29
add FTX.US and trade stream to FTX

vslee:
- refactored FTX to add FTX.US
- added ability to parse local DateTime in addition to UTC
- renamed Iso8601 enum to Iso8601UTC


jjxtra (approved):
Very good stuff!

PR #754, opened 2022-03-23
Set MarketSymbolIsUppercase to true in ExchangeBinanceAPI

### LINKED ISSUE #753: Binance's methods that have a symbol parameter throw an error.
I am using ExchangeSharp v1.0.0 .
Calling any method that takes in a symbol using an ExchangeAPI object connected to Binance throws the following exception.
For example, the following code:
```
await _api.GetTickerAsync("BTCUSDT");
```
Throws the following exception:
```
Unhandled exception. System.AggregateException: One or more errors occurred. ({"code":-1100,"msg":"Illegal characters found in parameter 'symbol'; legal range is '^[A-Z0-9-_.]{1,20}$'."})
---> ExchangeSharp.APIException: {"code":-1100,"msg":"Illegal characters found in parameter 'symbol'; legal range is '^[A-Z0-9-_.]{1,20}$'."}
```

This happens because the symbol's string is converted to lowercase somewhere along the way, and Binance does seem to require them to be in uppercase.
image

This shouldn't be hard to fix, I might do it myself if I find the time to understand where and why the string is converted.

miguel1117:
Fixes #753.
I set the MarketSymbolIsUppercase in the ExchangeBinanceAPI constructor and not in the BinanceGroupCommon constructor because I haven't tested if this change also applies to other Binance's exchanges.

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/BinanceGroup/ExchangeBinanceAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/BinanceGroup/ExchangeBinanceAPI.cs
@@ -18,6 +18,11 @@ namespace ExchangeSharp
 	{
 		public override string BaseUrl { get; set; } = "https://api.binance.com";
 		public override string BaseUrlWebSocket { get; set; } = "wss://stream.binance.com:9443";
+
+		private ExchangeBinanceAPI()
+		{
+			MarketSymbolIsUppercase = true;
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


vslee (approved):

vslee:
> I haven't tested if this change also applies to other Binance's exchanges.
 We can try it out for Binance proper first, and then other ppl can report if they need it for other exchanges

PR #766, opened 2022-04-09
new NuGet 1.0.1

vslee:
- added Coincheck and UPbit to README.md


jjxtra (approved):
Tip jar offer is still on the table... :)

vslee:
I appreciate the offer! What might be helpful is if you could look into moving this ES repo either into the DigitalRuby org or a new separate org. Then I can start looking into things like preview daily builds and SourceLink. Maybe even one day we can get it accepted into the .NET foundation.

jjxtra:
I can move it to digital ruby

vslee:
Great!

PR #788, opened 2022-09-23
Binance currencies null ref fix

jjxtra:


jjxtra:
I guess the asset code can be null...

vslee (approved):

PR #790, opened 2022-10-06
new NuGet 1.0.3

vslee:


jjxtra (approved):
Thanks for doing these @vslee

PR #801, opened 2023-04-01
Implementation of Bybit V5 unified API

TarasMi:
Implemented main private and public methods, some public (candles and orderbook) and private WebSockets. 
A decision was made to split spot, inverse, linear, and option branches into different APIs because: 
1) market symbols intersect, 
2) there is no method to get all the markets (all orders, all positions) from different branches in one call. Branches behave as separate exchanges.


vslee (approved):
Nice

PR #812, opened 2023-08-30
fix: Binance issue where Fees = null so null += Anything == null

O-Mutt:
This resolves an issue where the binance API returns the correct fees but the way we are handling fees is with a nullable decimal causing the += of fees to be null


O-Mutt:
I would hope we could get this released as a bug version as this is blocking my current production environment from being able to manage trades appropriately.

vslee (approved):

PR #826, opened 2024-01-23
chore: Adding better errors for coinbase order errors

O-Mutt:


src/ExchangeSharp/API/Exchanges/Coinbase/ExchangeCoinbaseAPI.cs
```diff
@@ -454,10 +454,22 @@ protected override async Task<ExchangeOrderResult> OnPlaceOrderAsync(ExchangeOrd
   			// The Post doesn't return with any status, just a new OrderId. To get the Order Details we have to reQuery.
   			return await OnGetOrderDetailsAsync(result[ORDERID].ToStringInvariant());
   		}
-  		catch (Exception ex) // All fails come back with an exception. 
+  		catch (Exception ex) // All fails come back with an exception.
   		{
+				Logger.Error(ex, "Failed to place coinbase error");
```
O-Mutt, line 459: Log the exception here

src/ExchangeSharp/API/Exchanges/Coinbase/ExchangeCoinbaseAPI.cs
```diff
@@ -454,10 +454,22 @@ protected override async Task<ExchangeOrderResult> OnPlaceOrderAsync(ExchangeOrd
...
+					IsBuy = payload["side"].ToStringInvariant().Equals(BUY),
+					MarketSymbol = payload["product_id"].ToStringInvariant(),
+					ClientOrderId = order.ClientOrderId,
+					ResultCode = $"{token["error_response"]["error"].ToStringInvariant()} - {token["error_response"]["preview_failure_reason"].ToStringInvariant()}",
+					AmountFilled = 0,
+					Amount = order.Amount,
+					AveragePrice = 0,
+					Fees = 0,
+					FeesCurrency = "USDT"
+				};
```
O-Mutt, line 472: include order details since the error handler uses `.ToString()` on the ExchangeOrderResults and without details it is useless. Also `token["error_response"]["error"]` seems insufficient in many cases so including `token["error_response"]["preview_failure_reason"]` as well

src/ExchangeSharp/API/Exchanges/Coinbase/ExchangeCoinbaseAPI.cs
```diff
@@ -396,24 +396,25 @@ protected override async Task<ExchangeOrderResult> OnPlaceOrderAsync(ExchangeOrd
   		payload["side"] = order.IsBuy ? BUY : "SELL";
 
   		Dictionary<string, object> orderConfig = new Dictionary<string, object>();
+			var amount = order.ShouldRoundAmount ? order.RoundAmount().ToStringInvariant() : order.Amount.ToStringInvariant();
```
O-Mutt, line 399: We check the ShouldRound and then `RoundAmount` if needed

O-Mutt: This check happens already in `order.RoundAmount` so i moved back to using the method instead of this variable check

thespragg (approved):

jjxtra (approved):

PR #829, opened 2024-02-07
fix: bug fix on buy order results

O-Mutt:
image

### DIFF
--- a/src/ExchangeSharp/API/Exchanges/OKGroup/ExchangeOKExAPI.cs
+++ b/src/ExchangeSharp/API/Exchanges/OKGroup/ExchangeOKExAPI.cs
@@ -831,34 +831,44 @@ namespace ExchangeSharp
 			);
 		}
 
-		private static ExchangeOrderResult ParseOrder(JToken token) =>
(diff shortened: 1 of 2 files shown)


src/ExchangeSharp/API/Exchanges/OKGroup/ExchangeOKExAPI.cs
```diff
@@ -831,34 +831,44 @@ await GetNoncePayloadAsync()
...
+			if (avgPrice == default && price != default)
+			{
+				avgPrice = price;
+			}
+			else if (price == default && avgPrice != default)
+			{
+				price = avgPrice;
+			}
+			newResult.Price = price;
+			newResult.AveragePrice = avgPrice;
```
O-Mutt, line 865: Seems there was an expectation that the inverse was true previously. I made it a bit safer parsing and use the other value if there is missing data one way or the other.

i.e. if we don't have avgPx use px, if we don't have px use avgPx, if we have both, Great!

jjxtra (approved):

